PR #5, opened 2023-04-16
Add ability to upload files to SQLite database

### LINKED ISSUE #4: `chel deploy`: add ability to deploy to sqlite
### Problem

We currently don't have the ability to support deploy to an sqlite3 db.

### Solution

Add this ability so that the signature is now:

```
chel deploy <url-or-dir-or-sqlitedb> <contract-manifest.json> [<manifest2.json> [<manifest3.json> ...]]
```

This way you can do:

```
chel deploy ./data/groupincome.db manifest.json
```

snowteamer:
Closes #4 

### Summary of changes
- `chel deploy` and `chel upload` now accept an SQLite database file as destination.
  - Currently, that database file must end with `.db` to be recognized.
  - It stills returns the file(s) hash(es) on success, but in the database `blob=${hash}` is used as key.
- A new `chel` Deno task to help in development. You can try any available `chel` command after a source file change without having to recompile the binary. For example, `deno task chel version` will rebuild `chel`'s JavaScript then run `chel version`.


vendor/import_map.json
```diff
@@ -1,9 +1,21 @@
...
+  "scopes": {
+    "./esm.sh/": {
+      "/v115/@multiformats/base-x@4.0.1/deno/base-x.mjs": "./esm.sh/v115/@multiformats/base-x@4.0.1/deno/base-x.mjs",
+      "/v115/multibase@4.0.6/deno/multibase.mjs": "./esm.sh/v115/multibase@4.0.6/deno/multibase.mjs",
+      "/v115/uint8arrays@3.1.1/deno/concat.js": "./esm.sh/v115/uint8arrays@3.1.1/deno/concat.js",
+      "/v115/uint8arrays@3.1.1/deno/from-string.js": "./esm.sh/v115/uint8arrays@3.1.1/deno/from-string.js",
+      "/v115/uint8arrays@3.1.1/deno/to-string.js": "./esm.sh/v115/uint8arrays@3.1.1/deno/to-string.js",
+      "/v115/varint@5.0.2/deno/varint.mjs": "./esm.sh/v115/varint@5.0.2/deno/varint.mjs",
+      "/v115/multiformats@9.9.0/deno/basics.js": "./esm.sh/v115/multiformats@9.9.0/deno/basics.js"
+    }
```
taoeffect, line 19: What is this new section? How did you generate it?

When I ran `deno task vendor` on `master` with the latest version of Deno (v1.32.4), it didn't add this section...

taoeffect: Looking [at the documentation](https://deno.land/manual@v1.32.4/basics/import_maps#overriding-imports), it seems like this is a section for creating some sort of overrides, but it doesn't seem like anything is being overwritten here, so this seems unnecessary?

snowteamer: I don't know, but in any case I didn't add this section manually. Now if I remove it, then run `deno task vendor`, it gets regenerated.

taoeffect: OK I guess we'll leave it then 🤷

src/database-sqlite.ts
```diff
@@ -0,0 +1,33 @@
...
+
+let db = null
+let databaseFilename
+let writeStatement
+
+export function checkKey (key: string): void {
+  if (/[/\\]/.test(key)) {
+    throw new Error(`bad key: ${key}`)
+  }
+}
```
taoeffect, line 13: Should this also be updated with the latest `checkKey` code from the other PR?

src/upload.ts
```diff
@@ -50,10 +54,20 @@ async function uploadToDir (filepath: string, dir: string) {
   return destination
 }
 
+async function uploadToSQLite (filepath: string, databaseFile: string) {
+  const { initStorage, writeData } = await import('./database-sqlite.ts')
+  initStorage(databaseFile)
+  const buffer = await Deno.readFile(filepath)
+  const hash = blake32Hash(buffer)
+  writeData(`blob=${hash}`, buffer)
```
taoeffect, line 62: It seems like this could result in a conflict between the two PRs?

If we're always storing files as blobs here, even if some of them aren't, then wouldn't that conflict with the way [`readData` is implemented here](https://github.com/okTurtles/group-income/pull/1527#pullrequestreview-1388983681)?

snowteamer: If we're always storing files as blobs here then `readData() `will have to be called used with `blob=` keys to get the data back. So the callers would need to be updated, rather than  `readData() ` itself I think.
We can also store files as text, but that will limit `chel deploy` to text files

taoeffect: Ok, thinking on this more, I think there is no conflict.

src/database-sqlite.ts
snowteamer: Yes, good catch

taoeffect (changes requested):
I think it would be useful if `chel deploy` could create the same DB that Chelonia creates in [this PR](https://github.com/okTurtles/group-income/pull/1527) if the database doesn't exist already. That way you don't need to worry about initializing it yourself when you run `chel deploy`, it's just instantly initialized for you.

src/upload.ts
taoeffect: @snowteamer Looking at it closer, https://github.com/okTurtles/group-income/pull/1552 does seem to not store with the prefix, so can we remove the prefix here?

taoeffect (changes requested):
I tried running `deno task vendor` to see how it works on my system, but get this error:

```
-> % deno task vendor
Task vendor deno vendor -f src/main.ts scripts/compile.ts
error: The source code is invalid, as it does not match the expected hash in the lock file.
  Specifier: https://esm.sh/blakejs@1.2.1
  Lock file: [path to]/chel/deno.lock
```

I tried to find out why it had the wrong hash for `blackjs`, but in a test project I couldn't even get it to generate a lock file with a hash for that dependency, so I'm quite confused. How did you generate the lock file? Feel free to reply on Slack.

taoeffect (changes requested):
It seems like this PR is generating the `deno.lock` file when `deno task build` gets run (or any `deno run` command), despite us using vendoring.

To fix:

1. Delete the `vendor/` folder and the `deno.lock` file (if it exists)
2. Delete the `"imports"` key `deno.json`
3. Run `deno task vendor`
4. Add `"https://deno.land/std/io/util.ts": "./deno.land/std@0.156.0/io/util.ts",` to `vendor/import_map.json`

Your import_map.json should now look like this:

```json
{
  "imports": {
    "https://esm.sh/multihashes@4.0.3?bundle": "./esm.sh/multihashes@4.0.3.js",
    "https://esm.sh/blakejs@1.2.1": "./esm.sh/blakejs@1.2.1.js",
    "https://deno.land/std/io/util.ts": "./deno.land/std@0.156.0/io/util.ts",
    "https://deno.land/": "./deno.land/",
    "https://esm.sh/": "./esm.sh/"
  }
}
```

5.

Push these changes.

taoeffect (approved):
Great work @snowteamer! I guess we're getting the hang of wrangling with Deno...

bin/chel
```diff
@@ -1 +0,0 @@
-placeholder, will be replaced by real binary
```
taoeffect, line 1: @snowteamer Just a heads up, since I missed this in my review. Removing this file was a mistake as it breaks the install.

PR #9, opened 2023-05-22
Add eventsSince command

### LINKED ISSUE #8: Implement "chel eventsSince" command
### Problem

`chel eventsSince` is a very useful command for interacting with the server and retrieving a list of events that can be inspected.

Say you want to know what happened in the group contract. `chel eventsSince` can be used to request a JSON array of all the events that happened on the `contractID` since the a particular hash. If the hash is the same as the `contractID`, it returns all the events, up to some limit that's specified by `--limit` that has a default of say 50 (and the server can add its own override on top of that).

### Solution

Implement this. In the README the type signature is defined as:

```
chel eventsSince [--limit N] <url> <contractID> <hash>
```

Here `<url>` refers to the Chelonia server. Ideally we'd be able to make this parameter `<remoteURL-or-localPath>`, allowing the command to work with both a remote server and a local database. It's up to whoever implements this if they want to do that for this issue or for a future followup.

snowteamer:
Closes #8

### DIFF
--- a/src/eventsSince.ts
+++ b/src/eventsSince.ts
@@ -17,7 +17,7 @@ const backends = {
 
 let backendFrom: typeof backends.sqlite | typeof backends.fs
 
-export async function eventsSince (args: string[]): Promise<string[]> {
+export async function eventsSince (args: string[]): Promise<void> {
   const parsedArgs = flags.parse(args)
 
   const limit = Number.parseInt(parsedArgs.limit ?? 50)
@@ -37,12 +37,13 @@ export async function eventsSince (args: string[]): Promise<string[]> {
 
   backendFrom = backends[from as keyof typeof backends]
   try {
-    await backendFrom.initStorage(from === 'fs' ? {dirname: src} : {dirname: path.dirname(src), filename: path.basename(src)})
+    const options = { internal: true, ...(from === 'fs' ? { dirname: src } : { dirname: path.dirname(src), filename: path.basename(src) }) }
+    await backendFrom.initStorage(options)
   } catch (error) {
     exit(`could not init storage backend at "${src}" to fetch events from: ${error.message}`)
   }
   const messages = await getMessagesSince(contractID, hash, limit)
-  console.log(Deno.inspect(messages.map(s => JSON.parse(s)), { colors: true, strAbbreviateSize: Infinity }))
+  console.log(JSON.stringify(messages.map(s => JSON.parse(s)), null, 2))
 }
 
 const getMessage = async function (hash: string): Promise<string> {
@@ -81,5 +82,5 @@ export async function getMessagesSince (contractID: string, since: string, limit
   } catch (error) {
     console.error(`[chel] ${error.message}:`, error)
   }
-  return entries
+  return entries.reverse()
 }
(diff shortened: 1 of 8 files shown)


taoeffect (changes requested):
Amazing @snowteamer! Well done! This is already coming in handy debugging stuff! 😄 👍 

One thing though: currently the events are being output with the newest at the top, and the oldest last. Can you reverse the order before printing?

src/eventsSince.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  backendFrom = backends[from as keyof typeof backends]
+  try {
+    await backendFrom.initStorage(from === 'fs' ? {dirname: src} : {dirname: path.dirname(src), filename: path.basename(src)})
+  } catch (error) {
+    exit(`could not init storage backend at "${src}" to fetch events from: ${error.message}`)
+  }
+  const messages = await getMessagesSince(contractID, hash, limit)
+  console.log(Deno.inspect(messages.map(s => JSON.parse(s)), { colors: true, strAbbreviateSize: Infinity }))
```
taoeffect, line 45: This looks great by itself, but it isn't parseable by the `jq` command because the output is not JSON. E.g. it displays objects and their keys are not wrapped in quotes.

We need to be able to run commands like this:

```sh
chel eventsSince groupincome.db 21XWnNJZQoeFQsmrSQm4FNKSVbErWHDsszBemXRgS1ZWPurSFs 21XWnNJZQoeFQsmrSQm4FNKSVbErWHDsszBemXRgS1ZWPurSFs | jq '.[] | select(.op[0] == "ae") | .op[1] | fromjson'
```

In order to do that this line needs to be:

```js
console.log(JSON.stringify(messages.map(s => JSON.parse(s)), null, 2))
```

taoeffect (changes requested):
One more change request!

Tried testing this with the `jq` command and ran into some issues parsing the output.

Also, in order for this command to be compatible with `jq` we need to comment out this line from `database-sqlite.ts`:

```js
console.log('Connected to the %s SQLite database.', filepath)
```

Otherwise `jq` chokes upon reading that line.

src/eventsSince.ts
snowteamer: Ok 👍

taoeffect (commented):
Oh! Can you add help documentation to `help.ts`?

snowteamer:
@taoeffect Ok, will try to write up something useful 🙂

src/eventsSince.ts
```diff
@@ -0,0 +1,102 @@
...
+  sqlite: await import('./database-sqlite.ts')
+}
+const defaultLimit = 50
+const headPrefix = 'head='
+
+export async function eventsSince (args: string[]): Promise<void> {
+  const parsedArgs = flags.parse(args)
+
+  const limit = Number(parsedArgs.limit ?? defaultLimit)
+  console.log(limit)
```
taoeffect, line 19: Oops, this breaks the output

snowteamer: Sorry I forgot to remove it

taoeffect (approved):
LGTM! Fantastic work @snowteamer! 👏 🎉

PR #10, opened 2023-05-31
Restrinct net and write permissions in conract-related commands

### LINKED ISSUE #1: Request perms at runtime to connect to net
### Problem

Need to make sure contract code doesn't try anything funny.

### Solution

https://deno.land/manual/runtime/permission_apis

Consider also doing the same for write permissions when `<url-or-dir>` is a directory that's outside of `.`

May want to allow net in general case (e.g. to make things like `deno deploy` smoother), and restrict it when using any of the contract-related commands.

See also comment in `src/upload.ts`

snowteamer:
Closes #1

### DIFF
--- a/src/upload.ts
+++ b/src/upload.ts
@@ -1,15 +1,12 @@
 'use strict'
 
 import { path, colors } from './deps.ts'
-import { blake32Hash, isDir } from './utils.ts'
+import { blake32Hash, isDir, revokeNet } from './utils.ts'
 
 // chel upload <url-or-dir-or-sqlitedb> <file1> [<file2> [<file3> ...]]
 
-// TODO: use Deno.permissions.request(...) to request permissions to the specific URL
-//       https://deno.land/manual/runtime/permission_apis
-//       and use this everywhere so that we protect against malicious contracts
-
 export async function upload (args: string[], internal = false) {
+  await revokeNet()
   const [urlOrDirOrSqliteFile, ...files] = args
   if (files.length === 0) throw new Error(`missing files!`)
   const uploaded = []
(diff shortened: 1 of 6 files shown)


src/upload.ts
```diff
@@ -1,15 +1,12 @@
...
+import { blake32Hash, isDir, revokeNet } from './utils.ts'
 
 // chel upload <url-or-dir-or-sqlitedb> <file1> [<file2> [<file3> ...]]
 
-// TODO: use Deno.permissions.request(...) to request permissions to the specific URL
-//       https://deno.land/manual/runtime/permission_apis
-//       and use this everywhere so that we protect against malicious contracts
-
 export async function upload (args: string[], internal = false) {
+  await revokeNet()
```
taoeffect, line 9: Uploading makes sense to have net permissions though. This function, `revokeNet`, is mainly useful for when we implement `latestState` I think

taoeffect (changes requested):
Nice work @snowteamer. Merge conflicts + 1 minor issue

src/upload.ts
taoeffect: Hmm. I did write

> See also comment in `src/upload.ts`

And I did write that comment. No idea why I wrote that in this file though 😅

snowteamer: Calling `revokeNet()` here ensures that the user gets a prompt so that only the upload URL gets allowed. Otherwise, no prompt would show up, but any URL would be implicitly allowed.

taoeffect: Yeah, but if we are using `chel upload` we expect a connection to be made. Might be annoying to have to confirm. Not only annoying, but if this is used as part of a build step, would be disruptive

taoeffect (approved):
LGTM! Great job @snowteamer!

PR #16, opened 2023-06-08
Add chel get command

### LINKED ISSUE #12: new command: chel get
### Problem

We need the following command and it doesn't exist:

```
chel get <url-or-folder-or-db> <key>
```

When using a URL, this queries the `GET /file/<key>` route.

### Solution

Implement the command and outputs the result to the console.

If it's a binary file, for example an image, the command will be used like this:

```
chel get https://url.com mygreatlongkey > file.png
```

snowteamer:
Closes #12

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,12 +1,19 @@
 'use strict'
 
-import { base58btc, blake, colors } from './deps.ts'
+import { base58btc, blake, colors, path } from './deps.ts'
 import { type Multibase } from './deps.ts'
+import * as fs from './database-fs.ts'
+import * as sqlite from './database-sqlite.ts'
+
+const backends = { fs, sqlite}
+
+export type Backend = typeof backends.sqlite | typeof backends.fs
 
 // TODO: implement a streaming hashing function for large files
 export function blake32Hash (data: string | Uint8Array): Multibase<'z'> {
   // TODO: for node/electron, switch to: https://github.com/ludios/node-blake2
   const uint8array = typeof data === 'string' ? new TextEncoder().encode(data) : data
+  // @ts-ignore Property 'blake2b256' does not exist on type '{}'.
   const digest = blake.blake2b.blake2b256.digest(uint8array)
   // While `digest.digest` is only 32 bytes long in this case,
   // `digest.bytes` is 36 bytes because it includes a multiformat prefix.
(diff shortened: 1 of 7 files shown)


vendor/import_map.json
```diff
@@ -1,8 +1,8 @@
 {
   "imports": {
-    "https://esm.sh/multiformats/bases/base58?pin=v120": "./esm.sh/multiformats@11.0.2/bases/base58.js",
-    "https://esm.sh/multiformats?pin=v120": "./esm.sh/multiformats@11.0.2.js",
-    "https://esm.sh/@multiformats/blake2?pin=v120": "./esm.sh/@multiformats/blake2@1.0.13.js",
+    "https://esm.sh/multiformats@11.0.2/bases/base58?pin=v120": "./esm.sh/multiformats@11.0.2/bases/base58.js",
+    "https://esm.sh/multiformats@11.0.2?pin=v120": "./esm.sh/multiformats@11.0.2.js",
+    "https://esm.sh/@multiformats/blake2@1.0.13?pin=v120": "./esm.sh/@multiformats/blake2@1.0.13.js",
```
taoeffect, line 5: Nice catch!

src/utils.ts
```diff
@@ -24,6 +31,26 @@ export function exit (message: string): never {
   Deno.exit(1)
 }
 
+export async function getBackend (src: string): Promise<Backend> {
```
taoeffect, line 34: This function seems to be only referenced from one place at the moment. Does it make sense to call it from more places to simplify code?

taoeffect (changes requested):
Nicely done @snowteamer! Question + request to fix merge conflict!

src/utils.ts
snowteamer: Yeah likely, but it's a bit complicated because this function only looks at the `src` string to find a backend whereas the command line for e.g. `chel migrate` specifies both the backend type e.g. `sqlite` and the source.

src/utils.ts
```diff
@@ -1,12 +1,19 @@
...
 
-import { base58btc, blake, colors } from './deps.ts'
+import { base58btc, blake, colors, path } from './deps.ts'
 import { type Multibase } from './deps.ts'
+import * as fs from './database-fs.ts'
+import * as sqlite from './database-sqlite.ts'
+
+const backends = { fs, sqlite}
+
+export type Backend = typeof backends.sqlite | typeof backends.fs
```
taoeffect, line 10: This is kinda cool 👍

src/utils.ts
taoeffect: I've opened up an issue related to this: https://github.com/okTurtles/chel/issues/20

Issue #11, opened 2023-05-31
Add a test suite


taoeffect:
### Problem

We don't have a test suite.

### Solution

Add a deno task command that uses Deno's built-in testing stuff (if it's viable) to test the `chel` command.

Add a dummy `groupincome.db` to the test assets and test that.

PR #21, opened 2023-08-06
Use CIDs for content addressing rather than plain hashes

### LINKED ISSUE #19: Add JSON encoding info to hash
### Problem

If we ever decide to change the encoding we use we should be able to without things breaking too badly.

### Solution

Add that info to the hash, using CIDs if necessary.

### LINKED ISSUE #20: Make more use of getBackend to DRY things
### Problem

In a few places in the code we repeat the concept of figuring out what backend we're dealing with.

### Solution

In those situations, use the new `getBackend` functions as described here: https://github.com/okTurtles/chel/pull/16#discussion_r1262832950

(Modifying things if necessary to DRY stuff)

snowteamer:
Closes #19
Closes #20

#### Content addressing
- `chel hash` now outputs [content identifiers (CIDs)](https://github.com/multiformats/cid) rather than unadorned hashes. For now, CIDs created by the command include the `raw` (0x55) code by default, and the `json` (0x0200) code for JSON files.
- It should now be much easier to switch the hashing algo or hash encoding, by updating constants.
- Improved `chel hash`  help message.

#### Testing (related to #11)
- Added tests for `chel hash`.
- Added tests for `getBackend()`.
- New `deno task test` command, to run the test suite with suitable CLI flags.
- Files under `./test/assets/` are no longer git-ignored.

#### Database backends
- Added a read-only `dataFolder` property to database backends. Useful e.g. in tests.
- Fixed an error when calling `initStorage()` more than once on the `sqlite` backend.

#### Other
- Refactored `chel eventsAfter` and `chel migrate` using the new `getBackend()` helper (related to #20)
- Added a minimal `.vscode/settings.json` file to make sure the Deno extension is enabled in VS Code / Codium.
- Fixed missing `--no-remote` flag in `deno task chel`.
- Made the `isDir()` and `isFile()` utilities async.


src/utils.ts
```diff
@@ -1,31 +1,46 @@
...
+}
+
+// TODO: implement a streaming hashing function for large files.
+// Note: in fact this returns a serialized CID, not a CID object.
+export function createCID (data: string | Uint8Array, multicode = multicodes.RAW): string {
+  const uint8array = typeof data === 'string' ? new TextEncoder().encode(data) : data
+  const digest = multihasher.digest(uint8array)
+  const cid = CID.create(1, multicode, digest)
+  const key = cid.toString(multibase.encoder)
+  checkKey(key)
```
taoeffect, line 40: Don't think this check needs to be here

snowteamer: Yeah, not necessary in theory, I just thought it was a nice additional safety since this code is still new

src/database-fs.ts
```diff
@@ -4,7 +4,7 @@ import { path } from './deps.ts'
 import { checkKey } from './utils.ts'
 
 // Initialized in `initStorage()`.
-let dataFolder = ''
+export let dataFolder = ''
```
taoeffect, line 7: Hmm, why is this being exported? 🤔 

I don't see it imported anywhere... and isn't kind of a weird thing to do, to export such a variable?

src/utils.ts
```diff
@@ -1,76 +1,103 @@
...
+  return path.slice(index).toLowerCase()
+}
+
 export function isArrayLength (arg: number): boolean {
   return Number.isInteger(arg) && arg >= 0 && arg <= 2 ** 32 - 1
 }
 
 // Checks whether a path points to a directory, following symlinks if any.
-export function isDir (path: string | URL): boolean {
+export async function isDir (path: string | URL): Promise<boolean> {
```
taoeffect, line 88: Why did you change these to be async?

And if there's a strong reason to, can you add that as a comment?

vendor/binary-install/index.js
```diff
@@ -1,106 +0,0 @@
-// we directly include this in our project so that the downloaded binary is stored
-// with our module instead of with some other one. This allows us to avoid using
-// nodejs as a wrapper executable to run our binary, greatly speeding up performance
-// see: https://github.com/EverlastingBugstopper/binary-install/tree/main/packages/binary-install
-
-const { existsSync, mkdirSync } = require("fs")
```
taoeffect, line 6: Why did you delete this file?

I'm pretty sure this file must not be deleted or it will break the install

taoeffect (changes requested):
Very nicely done @snowteamer! Left a few change requests + questions

src/hash.ts
```diff
@@ -12,13 +11,9 @@ export async function hash (args: string[], internal = false) {
...
   }
-  // TODO: implement this in a way that chunks the input, so that we don't read the entire file into memory at once
-  const file = await Deno.open(filename, {read: true})
-  const myFileContent = await streams.readAll(file)
-  Deno.close(file.rid)
-  const hash = blake32Hash(myFileContent)
+  const [cid] = await createEntryFromFile(filename)
   if (!internal) {
-    console.log(`blake32Hash(${filename}):`, hash)
+    console.log(`CID for file (${filename}):`, cid)
```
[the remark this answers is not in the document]
snowteamer: Will do!

src/utils.ts
snowteamer: Nowadays async is generally the default way for any I/O operation in JavaScript, and the fact these functions were sync caught my eye. Agreed, not sure whether there is a real performance benefit in the CLI tool use case, at least as long as the functions aren't called in loops. But in principe it still allows the engine to not forcibly stay idle while waiting for the operation to complete. And we might start using these functions in loops sooner or later, in which case being async enables usage of e.g. `Promise.all()`. 
Also these function names have no `-Sync` suffix unlike native Nodejs and Deno sync IO functions

taoeffect: Alright 👍

src/database-fs.ts
snowteamer: > I don't see it imported anywhere...

It's available as `backend.dataFolder` once the whole module has been imported as `backend`. It's not meant to be imported separately.

> why is this being exported?

We might not need this in common use cases, but in some `getBackend` tests where the same backend object is "recycled" to use a new data folder, I could not easily make sure it had been correctly updated without having access to this propery.

Of course this happens because in this design, the backend object is the result of exporting its whole module, like `export * as fs from './src/database-fs.ts'`. In other words, the module acts as a singleton instance in OOP terminology, in which case exporting a module variable is like making a property of that singleton public rather than keeping it private.
So I guess it isn't that weird in the context of that design. But maybe that design itself is a bit weird.

By the way, the ES module spec ensures an exported `let` binding is read-only outside its module (although mutable from within it)

test/hash.test.ts
```diff
@@ -0,0 +1,36 @@
...
+            const actual = await hash([filepath], true)
+            const expected = 'zyop8PQypgycyyCEcQsEHoc9Z8S9bQt1CuHuRKnXrAXagVBGxQyvH'
+            assertEquals(actual, expected)
+
+            const cid = CID.parse(actual, base58btc.decoder)
+            assertEquals(cid.code, 0x0200)
+            assertEquals(cid.byteLength, 39)
+            assertEquals(cid.multihash.code, 45600)
+            assertEquals(cid.version, 1)
+        })
```
taoeffect, line 34: 👏 Bravo on these tests btw!!

This is definitely very much needed for sanity checking!!

taoeffect (approved):
LGTM!

Issue #26, opened 2024-02-08
Implement signing manifests


taoeffect:
### Problem

We currently do not have contract signatures implemented for the `chel manifest` command.

### Solution

1. Implement [`chel keygen`](https://github.com/okTurtles/chel?tab=readme-ov-file#chel-keygen)
2. Add signing to [`chel manifest`](https://github.com/okTurtles/chel?tab=readme-ov-file#chel-manifest)

If instead of `"authors"` you prefer `"signingKeys"` for the attribute name, you can use that instead.

taoeffect:
Closed by @corrideat in #27

PR #27, opened 2024-02-08
Implemented manifest signatures

corrideat:
Closes issue #26


src/keygen.ts
```diff
@@ -0,0 +1,21 @@
...
+    version: '1.0.0',
+    pubkey: serializeKey(key, false),
+    privkey: serializeKey(key, true)
+  }))
+
+  if (parsedArgs['out']) {
+    await Deno.writeFile(parsedArgs['out'], result)
+  } else {
+    await Deno.stdout.write(result)
+  }
```
taoeffect, line 20: Request:

- output 2 files, one for the public key (for verifying with `chel latestState`), and one for the private key for signing with `chel manifest` (like SSH)
- remove the STDOUT. you can have a default filename based on the key type (make sure to output to STDOUT where this file is saved, and check to see if it already exists, in which case, just add a suffix like `id_edssa.1.pub` and `id_edssa.2.priv`) or something to that effect

taoeffect (changes requested):
Great work @corrideat!!

I left some feedback and also realized we need two tests for this:

1. A test to verify that contracts are signed correctly using a new `chel verifyContract` command that takes a pubkey file and a manifest (can update one of the manifest files in `test/data/`)
2. A test to verify that a modified contract (after it's been signed) fails a signature check (can use the `test/data/mailbox.*` files for this one)

src/keygen.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const idx = keyId(key).slice(-12)
+  const outFile = parsedArgs['out'] || `${EDWARDS25519SHA512BATCH}-${idx}.json`
+  const pubOutFile = parsedArgs['pubout'] || `${EDWARDS25519SHA512BATCH}-${idx}.pub.json`
+
+  await Deno.writeTextFile(outFile, result)
+  console.log(colors.green('wrote:'), outFile, colors.green('(secret)'))
+
+  await Deno.writeTextFile(pubOutFile, pubResult)
+  console.log(colors.green('wrote:'), pubOutFile, colors.green('(public)'))
```
taoeffect, line 28: Can you make the color for "(secret)" and "(public)" blue or grey?

src/verifySignature.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const body = JSON.parse(manifest.body)
+  if (!Array.isArray(body.signingKeys)) throw new Error('Missing signingKeys in manifest')
+  if (body.signingKeys.every((k: string) => {
+    return keyId(k) !== id
+  })) {
+    throw new Error('The signing key is not listed in signingKeys')
+  }
+
+  if (!internal) console.log(colors.green('ok'), 'all checks passed')
```
taoeffect, line 49: - If the keyfile isn't specified in the arguments, see if the signature passes one of the keys in the manifest
- If the keyfile is specified, check if the signature passes that key only, and *IF* that key is not also specified in the manifest, then output a yellow bold `console.warn` message saying that the check passed the keyfile but that key wasn't found on in the manifest

src/manifest.ts
```diff
@@ -7,42 +7,62 @@
...
   const manifest = JSON.stringify({
-    head: { manifestVersion: "1.0.0" },
-    body: JSON.stringify(body),
+    head,
+    body: serializedBody,
     signature: {
-      key: "<which of the 'authors' keys was used to sign 'body'>",
-      signature: "<signature>"
+      keyId: keyId(signingKey),
+      value: sign(signingKey, serializedBody + head.manifestVersion)
```
taoeffect, line 65: Can you sign full head?

src/verifySignature.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  if (!manifest.signature) {
+    exit('Invalid manifest: missing signature', internal)
+  }
+  if (!manifest.signature.keyId) {
+    exit('Invalid manifest: missing signature key ID', internal)
+  }
+  if (!manifest.signature.value) {
+    exit('Invalid manifest: missing signature value', internal)
+  }
```
taoeffect, line 32: We also need to load both contract files (slim and bundle) and verify the hash matches the hash that's in the manifest.

taoeffect (approved):
Perfect! 🎉🎉

PR #41, opened 2025-01-05
Use correct CIDs

corrideat:


src/deploy.ts
```diff
@@ -13,11 +13,11 @@ export async function deploy (args: string[]) {
     const json = JSON.parse(Deno.readTextFileSync(manifestPath))
     const body = JSON.parse(json.body)
     const dirname = path.dirname(manifestPath)
-    toUpload.push(path.join(dirname, body.contract.file))
+    toUpload.push('t|' + path.join(dirname, body.contract.file))
```
taoeffect, line 16: What is this `t|` and `m|` stuff? Can you add a comment explaining?

corrideat: It's a prefix because we need to know what kind of CID should be used.

src/upload.ts
```diff
@@ -14,8 +14,26 @@ export async function upload (args: string[], internal = false) {
...
+          break
+        case 't':
+          type = multicodes.SHELTER_CONTRACT_TEXT
+          break
+        default:
+          throw new Error('Unknown file type: ' + filepath_[0])
+      }
+      filepath = filepath_.slice(2)
+    }
+    const entry = await createEntryFromFile(filepath, type)
```
taoeffect, line 36: Is it not a concern that a user might use `chel upload` on a file that has `|` as the 2nd character?

corrideat: Files can't have `|` on Windows (one of the reasons I chose it. However, for such files, they could be prefixed as needed. E.g., if the filename is `||`, one could use `r|||`

corrideat: On POSIX, only `0x00` (and maybe `/`) is an invalid character in file name, but using `0x00` for this purpose looked wrong, as it's difficult to type in a terminal.

taoeffect: OK, then do you make sure that everywhere `upload` is called the passed in files are prefixed?

And if they're all prefixed, what's the point of check on line 20? Instead, we should probably throw an exception if `|` isn't found.

corrideat: The intent was to allow the command to be used from the command line without a prefix (in which case a raw id would be used)

taoeffect: I think if it's used from the command line we can detect that by checking the `internal` variable, right?

If so, then you should be able to throw an exception when `internal = true` and we don't have `|`, and when `internal = false` we can add the prefix manually inside of the `upload` function before the `switch` statement hits.

corrideat: But what if the user wishes to set the prefix themselves?

taoeffect: Then we can invent some other flag or function for that.

corrideat: But a flag wouldn't work very well because the `upload` command works with multiple files.

corrideat: Sure, but I don't really see the ambiguity here as `|` is pretty rare as a file name, and I expect that having a file called `r|XX` and another one called `XX` is even rarer.

taoeffect: I agree, it's rare, but not so rare that it never happens. On macOS `|` is a valid filename character. If we can fix this now and forget about it, we don't need to ever worry about it in the future.

taoeffect (changes requested):
Yeah I think just eliminating all chance of any confusion over the filename prefixes and this PR is good to merge 👍

src/upload.ts
```diff
@@ -14,8 +14,28 @@ export async function upload (args: string[], internal = false) {
...
```
taoeffect, line 38: This `filepath` here seems out of scope as it was defined inside the loop?

taoeffect: 🤦‍♂️

taoeffect (approved):
Nice!

Issue #42, opened 2025-06-30
Deno 2 compatibility


taoeffect:
### Problem

`deno task dist` doesn't work with Deno 2.

### Solution

Fix all errors.

Many of the errors are linter errors linting `test/assets/group.js` and related test files. These errors can be addressed simply by telling Deno to ignore those files.

taoeffect:
Closed by @akhileshthite in #46!

PR #44, opened 2025-07-04
fix: adapt chel CLI for Deno 2 linting and environment compatibility

akhileshthite:
Closes https://github.com/okTurtles/chel/issues/42

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -5,7 +5,7 @@
     "vendor": "deno vendor -f src/main.ts scripts/compile.ts",
     "compile": "deno run --import-map=./vendor/import_map.json --no-remote --allow-run --allow-read=. --allow-write=./dist scripts/compile.ts",
     "build": "deno run --import-map=./vendor/import_map.json --no-remote --allow-run --allow-read --allow-env --allow-write=./build --allow-net scripts/build.ts",
-    "dist": "deno lint && deno task build && deno task compile",
+    "dist": "deno lint . && deno task build && deno task compile",
     "test": "deno test --import-map=./vendor/import_map.json --allow-read=. --allow-write=test --no-remote"
   },
   "imports": {
@@ -21,9 +21,7 @@
     "singleQuote": true
   },
   "lint": {
-    "files": {
-      "exclude": ["node_modules/", "test/", "vendor/", "dist/", "build/"]
-    }
+    "exclude": ["node_modules/", "test/", "test/assets/group.js", "test/assets/mailbox.js", "vendor/", "dist/", "build/"]
   },
   "importMap": "./vendor/import_map.json"
 }
--- a/deno.json
+++ b/deno.json
@@ -21,7 +21,7 @@
     "singleQuote": true
   },
   "lint": {
-    "exclude": ["node_modules/", "test/", "scripts/binary.js", "test/assets/group.js", "test/assets/mailbox.js", "vendor/", "dist/", "build/"]
+    "exclude": ["node_modules/", "test/", "test/assets/group.js", "test/assets/mailbox.js", "vendor/", "dist/", "build/"]
   },
   "importMap": "./vendor/import_map.json"
 }
--- a/deno.json
+++ b/deno.json
@@ -21,7 +21,7 @@
     "singleQuote": true
   },
   "lint": {
-    "exclude": ["node_modules/", "test/", "test/assets/group.js", "test/assets/mailbox.js", "vendor/", "dist/", "build/"]
+    "exclude": ["node_modules/", "vendor/", "dist/", "build/", "test/assets/**/*.js"]
   },
   "importMap": "./vendor/import_map.json"
 }
--- a/deno.json
+++ b/deno.json
@@ -21,7 +21,7 @@
     "singleQuote": true
   },
   "lint": {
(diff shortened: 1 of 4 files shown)


src/version.ts
```diff
@@ -1,6 +1,5 @@
-'use strict'
```
taoeffect, line 1: Is there a reason this was removed?

corrideat:
It’s not really needed.

Den fre., 4 juli, 2025 kl. 04:06, Greg Slepak ***@***.***(mailto:Den fre., 4 juli, 2025 kl. 04:06, Greg Slepak <<a href=)> skrev:

taoeffect (changes requested):
Nice! Thanks for sending in the PR, looks like it's almost working, but on my machine with Deno 2.3.1 I'm running into this issue when running `deno task dist`:

```
-> % deno task dist
Warning "importMap" field is ignored when "imports" or "scopes" are specified in the config file.
    at chel/deno.json
Task dist deno lint . && deno task build && deno task compile
Warning "importMap" field is ignored when "imports" or "scopes" are specified in the config file.
    at chel/deno.json
Unsupported compiler options in "chel/deno.json".
  The following options were ignored:
    allowJs
Checked 24 files
Warning "importMap" field is ignored when "imports" or "scopes" are specified in the config file.
    at chel/deno.json
Task build deno run --import-map=./vendor/import_map.json --no-remote --allow-run --allow-read --allow-env --allow-write=./build --allow-net scripts/build.ts
Warning "importMap" field is ignored when "imports" or "scopes" are specified in the config file.
    at chel/deno.json
Warning the configuration file "chel/deno.json" contains an entry for "importMap" that is being ignored.
Unsupported compiler options in "chel/deno.json".
  The following options were ignored:
    allowJs
✘ [ERROR] Could not resolve "node:process"

    src/version.ts:1:20:
      1 │ import process from "node:process"
        ╵                     ~~~~~~~~~~~~~~

  You can mark the path "node:process" as external to exclude it from the bundle, which will remove
  this error.

error: Uncaught (in promise) Error: Build failed with 1 error:
src/version.ts:1:20: ERROR: Could not resolve "node:process"
  let error = new Error(`${text}${summary}`);
```

Also left a question for you!

deno.json
```diff
@@ -21,9 +21,7 @@
     "singleQuote": true
   },
   "lint": {
-    "files": {
-      "exclude": ["node_modules/", "test/", "vendor/", "dist/", "build/"]
-    }
+    "exclude": ["node_modules/", "test/", "scripts/binary.js", "test/assets/group.js", "test/assets/mailbox.js", "vendor/", "dist/", "build/"]
```
corrideat, line 24: Is it not possible to use glob expressions for things in `test/assets`? Also, why is `binary.js` being excluded? It doesn't look like it should be

src/version.ts
akhileshthite: Deno enforces strict mode automatically https://github.com/denoland/deno/issues/5845

deno.json
akhileshthite: Got `NodeJS process global is discouraged in Deno` for `binary.js`

The script is classic Node CLI script so instead of ES modules refactor I thought it's best to ignore it.

taoeffect:
> I'm on 2.4.0

It has nothing to do with the Deno version (just verified on 2.4.0). 

> We can also ignore version.ts?

That file needs to be checked just like all the other ones.

BTW, also am noticing a problem with `deno task lint`...

```
deno task lint
Warning "importMap" field is ignored when "imports" or "scopes" are specified in the config file.
    at chel/deno.json
Task lint standard ./src
standard: command not found
```

That should probably be fixed too.

akhileshthite:
> > I'm on 2.4.0
> 
> It has nothing to do with the Deno version (just verified on 2.4.0).
> 
> > We can also ignore version.ts?
> 
> That file needs to be checked just like all the other ones.
> 
> BTW, also am noticing a problem with `deno task lint`...
> 
> ```
> deno task lint
> Warning "importMap" field is ignored when "imports" or "scopes" are specified in the config file.
>     at chel/deno.json
> Task lint standard ./src
> standard: command not found
> ```
> 
> ~That should probably be fixed too.~ EDIT: let's save fixing that for a separate issue.
> 
> Opened a separate issue for that: #45

Yeah, I was gonna open issue for that.
Thanks!

corrideat:
> You can mark the path "node:process" as external to exclude it from the bundle, which will remove
  this error.

That sounds like it would work. It is an external module after all.

deno.json
corrideat: Also: the glob comment also applies to the other directories. It `test` didn't work before, the other directories also won't work

akhileshthite:
> ✘ [ERROR] Could not resolve "node:process"
> 
>     src/version.ts:1:20:
>       1 │ import process from "node:process"
>         ╵                     ~~~~~~~~~~~~~~
> 
>   You can mark the path "node:process" as external to exclude it from the bundle, which will remove
>   this error.

`external: ['node:process']` in `scripts/build.ts`?

taoeffect:
From the logs it says that script/build.ts is being used, which uses esbuild, so yes I'd look at the API docs here: https://esbuild.github.io/

akhileshthite:
All the commits are present in https://github.com/okTurtles/chel/pull/46 since that branch was created from this fork.

PR #46, opened 2025-07-05
fix: resolve vendor import issues and update dependency handling for deno 2.x

akhileshthite:
This PR resolves a series of issues related to vendor directory imports and dependency handling in a Deno 2.x environment, ensuring a smooth build and compilation process for the Chel CLI.

### deno.json: 
- Emptied imports to remove all mappings, relying on direct imports in code to avoid import map-related warnings.
- Emptied `compilerOptions` to ensure no unused options trigger warnings.
- Retained `tasks.build` with --node-modules-dir and removed `miniexec` (why? `miniexec` was that its outdated Deno.run() and improper `Deno.Command` usage caused stream handling errors, leading to vendor import issues when bundled.)
- Added `"vendor": true` as per deno 2; `deno task vendor` is deprecated.

### src/deps.ts:
- Removed `export { miniexec as sh } from "https://deno.land/x/miniexec@1.0.0/mod.ts";` and replaced with local import structure.

### scripts/compile.ts:
- Updated `import { sh } from '../src/deps.ts'` to `import { shell } from '~/utils.ts'`.

### scripts/build.ts:
- Added JSR imports in `external`.

### vendor/import_map.json:
- Added mappings for ESM and JSR dependencies and removed the deprecated ones.

### Working Commands
✅ deno task chel
✅ deno task build
✅ deno task test
✅ deno task lint
✅ deno task dist (partially, fails at compile step)
❌ deno task compile:
Error:
```
Task compile deno run --allow-run --allow-read=. --allow-write=./dist scripts/compile.ts
./dist/tmp
./dist/tmp/x86_64-unknown-linux-gnu
error: Importing from the vendor directory is not permitted. Use a remote specifier instead or disable vendoring.
    at /build/main.js:27:21
caught: Command failed with exit code 1: error: Importing from the vendor directory is not permitted. Use a remote specifier instead or disable vendoring.
    at f/build/main.js:27:21
```
If `"vendor": true` removed from `deno.json`:
Error:
```
Task compile deno run --allow-run --allow-read=. --allow-write=./dist scripts/compile.ts
./dist/tmp
./dist/tmp/x86_64-unknown-linux-gnu
error: JSR package manifest for '@std/assert' failed to load. A remote specifier was requested: "https://jsr.io/@std/assert/meta.json", but --no-remote is specified.
    at /build/main.js:14:67
caught: Command failed with exit code 1: error: JSR package manifest for '@std/assert' failed to load. A remote specifier was requested: "https://jsr.io/@std/assert/meta.json", but --no-remote is specified.
    at /build/main.js:14:67
```


src/deps.ts
```diff
@@ -1,21 +1,18 @@
...
+export { copy, readAll } from 'jsr:@std/io';
+export { default as tweetnacl } from 'https://esm.sh/tweetnacl@1.0.3?pin=v120';
+export { base58btc } from 'https://esm.sh/multiformats@11.0.2/bases/base58?pin=v120';
+export { type Multibase } from 'https://esm.sh/multiformats@11.0.2?pin=v120';
+export { default as blake } from 'https://esm.sh/@multiformats/blake2@1.0.13?pin=v120';
+export { CID } from 'https://esm.sh/multiformats@11.0.2/cid?pin=v120';
+// export { miniexec as sh } from "https://deno.land/x/miniexec@1.0.0/mod.ts";
+export * as esbuild from "https://deno.land/x/esbuild@v0.14.47/mod.js";
+export * as sqlite from "https://deno.land/x/sqlite@v3.7.1/mod.ts";
+export { type DB as SQLiteDB, type PreparedQuery } from "https://deno.land/x/sqlite@v3.7.1/mod.ts";
```
akhileshthite, line 18: @taoeffect 

The reason for these remaining deno.land imports (esbuild and sqlite) are retained because I think they are essential dependencies for the build and database functionality. But I'm happy to use any alternatives.

deno.json
```diff
@@ -1,29 +1,32 @@
...
+    "~/": "./src/",
+    "jsr:@std/": "https://jsr.io/@std/",
+    "https://esm.sh/tweetnacl@1.0.3?pin=v120": "https://esm.sh/tweetnacl@1.0.3?pin=v120",
+    "https://esm.sh/multiformats@11.0.2/bases/base58?pin=v120": "https://esm.sh/multiformats@11.0.2/bases/base58?pin=v120",
+    "https://esm.sh/multiformats@11.0.2?pin=v120": "https://esm.sh/multiformats@11.0.2?pin=v120",
+    "https://esm.sh/@multiformats/blake2@1.0.13?pin=v120": "https://esm.sh/@multiformats/blake2@1.0.13?pin=v120",
+    "https://esm.sh/multiformats@11.0.2/cid?pin=v120": "https://esm.sh/multiformats@11.0.2/cid?pin=v120",
+    "https://deno.land/x/miniild@v0.14.47/mod.js": "https://deno.land/x/esbuild@v0.14.47/mod.js",
+    "https://deno.land/x/sqlite@v3.7.1/mod.ts": "https://deno.land/x/sqlite@v3.7.1/mod.ts"
   },
```
akhileshthite, line 20: @taoeffect 
The imports field in `deno.json` provides a global mapping for the project to resolve dependencies like JSR and ESM modules, while the `import_map.json` offers a more detailed, file-specific mapping for offline or vendored use.

We can have both, I removed `import_map.json` because it was giving a lot of warnings. Also thought having both can lead to conflicts because I've also added JSR in `import_map.json`.

src/deps.ts
taoeffect: I'm not suggesting you use alternatives. AFAICT, you weren't using alternatives for the other deno.land => jsr conversions you made... I'm asking why you aren't loading these from jsr too?

taoeffect: Ah, yeah feel free to try with more recent versions then 👍

deno.json
akhileshthite: Oh I meant I removed `import_map.json` from `deno.json` to avoid warnings.

You're right; most of these mappings are redundant as they don’t alter the original URLs, so we can remove them to simplify `deno.json`.

scripts/compile.ts
```diff
@@ -15,7 +15,7 @@ export async function compile () {
     const bin = arch.includes('windows') ? 'chel.exe' : 'chel'
     // note: could also use https://examples.deno.land/temporary-files
     await $(`mkdir -vp ${dir}`)
-    await $(`deno compile --allow-read=./ --allow-write=./  --allow-net --no-remote --import-map=vendor/import_map.json -o ${dir}/${bin} --target ${arch} ./build/main.js`)
```
akhileshthite, line 18: Removed remote import and JSON import since build is used directly.

deno.json
```diff
@@ -20,6 +21,7 @@
...
   "compilerOptions": {
   },
+  "nodeModulesDir": "auto",
```
akhileshthite, line 17: To allow Deno to auto-resolve and embed necessary Node packages.

deno.json
```diff
@@ -10,6 +10,7 @@
   "imports": {
     "~/": "./src/",
     "jsr:@std/": "https://jsr.io/@std/",
+    "npm:@types/node": "npm:@types/node@latest",
```
akhileshthite: To support Node-style typings and compatibility with tools that expect it.

src/deps.ts
taoeffect: @akhileshthite These still seem to be using deno.land...?

scripts/compile.ts
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env -S deno run --allow-run --allow-read=. --allow-write=./dist
 
-import { sh } from '../src/deps.ts'
+import { shell } from '~/utils.ts';
```
taoeffect, line 3: What is the reason you decided to not use the dependency?

scripts/build.ts
```diff
@@ -12,6 +12,17 @@ const options = {
...
+    'node:process',
+    'jsr:@std/assert',
+    'jsr:@std/encoding/base64',
+    'jsr:@std/flags/',
+    'jsr:@std/fmt/colors',
+    'jsr:@std/fs/',
+    'jsr:@std/path/',
+    'jsr:@std/streams/',
+    'jsr:@std/io'
+  ],
```
taoeffect, line 25: Hmm... I think we will need to double-check that this doesn't break the distributed binary by excluding these from the `build/main.ts` file...

taoeffect (changes requested):
Nice work @akhileshthite, review is ready!

I also replied to update an old comment from last week that was unaddressed, so please check all threads.

scripts/compile.ts
akhileshthite: Yeah, I removed `miniexec`; its outdated Deno.run() and improper Deno.Command usage caused stream handling errors, leading to vendor import issues when bundled. Also couldn't find `jsr:` for it.

So just copied this simple `sh` function from `miniexec` and moved into `utils` as `shell`: https://github.com/okTurtles/chel/blob/038b71bcdfb64def7688e896917f52bbc0a0ec16/src/utils.ts#L150

build/main.js
```diff
@@ -964,8 +965,9 @@ var verifySignature2 = async (args, internal = false) => {
 };
 
 // src/version.ts
+import process from "node:process";
 function version() {
-  console.log("3.0.0");
+  console.log(process.env.VERSION);
```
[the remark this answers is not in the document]
akhileshthite: Thanks for noticing, fixed it with https://github.com/okTurtles/chel/pull/46/commits/ccc418a178a3fb022744dd7bd6344a791959d4fe

> Ignore the `vendor/` changes, that's because I removed the `dir` and ran `deno task dist`

scripts/build.ts
akhileshthite: The `external` modules are required for the build to succeed due to JSR resolution issues.
The distributed binary works offline (`deno run build/main.js`) even after clearing the cache with `deno cache --reload src/deps.ts`, suggesting the modules are bundled rather than relying on cached runtime resolution.

src/deps.ts
```diff
@@ -1,21 +1,17 @@
...
+export * as util from 'jsr:@std/io';
+export { copy, readAll } from 'jsr:@std/io';
+export { default as tweetnacl } from 'https://esm.sh/tweetnacl@1.0.3?pin=v120';
+export { base58btc } from 'https://esm.sh/multiformats@11.0.2/bases/base58?pin=v120';
+export { type Multibase } from 'https://esm.sh/multiformats@11.0.2?pin=v120';
+export { default as blake } from 'https://esm.sh/@multiformats/blake2@1.0.13?pin=v120';
+export { CID } from 'https://esm.sh/multiformats@11.0.2/cid?pin=v120';
+export * as sqlite from "jsr:@db/sqlite";
+export { type Database as SQLiteDB } from "jsr:@db/sqlite";
+export * as esbuild from "https://deno.land/x/esbuild@v0.25.6/mod.js";
```
taoeffect, line 17: Is `esbuild` not available on `jsr`?

src/deps.ts
```diff
@@ -1,21 +1,17 @@
...
-// https://deno.land/x/miniexec
+export { assert, assertEquals, assertRejects, assertThrows } from 'jsr:@std/assert';
+export * as base64 from 'jsr:@std/encoding/base64';
+export * as flags from 'jsr:@std/flags/';
+export * as colors from 'jsr:@std/fmt/colors';
+export * as fs from 'jsr:@std/fs/';
+export * as path from 'jsr:@std/path/';
+export * as streams from 'jsr:@std/streams/';
+export * as util from 'jsr:@std/io';
+export { copy, readAll } from 'jsr:@std/io';
```
corrideat, line 9: The previous import lines pinned a specific version, while the new imports don't seem to do that. Can you please use a specific version?

deno.json
```diff
@@ -1,29 +1,27 @@
 {
   "tasks": {
-    "chel": "deno task build && deno run --import-map=./vendor/import_map.json --allow-net --allow-read=. --allow-write=. --no-remote build/main.js",
+    "chel": "deno task build && deno run --allow-net --allow-read=. --allow-write=. build/main.js",
     "lint": "standard ./src",
```
corrideat, line 4: See my 'review' comment about this specific line. Also, an advantage of using `npx` is that a specific version can be pinned.

deno.json
```diff
@@ -1,29 +1,27 @@
...
   "fmt": {
     "lineWidth": 100,
     "semiColons": false,
     "singleQuote": true
   },
   "lint": {
-    "files": {
-      "exclude": ["node_modules/", "test/", "vendor/", "dist/", "build/"]
-    }
+    "exclude": ["node_modules/", "vendor/", "dist/", "build/", "test/assets/**/*.js"]
```
corrideat, line 24: I don't see a need for excluding `js` files in particular. `test/assets/` should be good enough, since no files whatsoever under this directory should be linted.

akhileshthite:
> I've noticed two issues with this repository as is (which maybe existed from before):
> 
> 1. I can't run `deno task lint`. It says `standard: command not found`. I don't have `standard` in my path, nor do I think I should. Changing this to `npx standard` fixes it.
> 2. Even after applying the above fix, the invocation is wrong. It should be `(npx)` `standard ./src/**`
> 3. Even after running `standard ./src/**`, there are several standard violations.
> 
> ETA: Most (all?) of the errors when running `standard` seem to be due to parse errors, since the source files are TypeScript. Using `ts-standard` would address this, except that we then still end up with lots of linting errors. @taoeffect thoughts?

Yes exactly, I've used `ts-standard` for linting.. I'm sending a separate PR for that.

src/utils.ts
```diff
@@ -138,4 +138,40 @@ export async function revokeNet () {
...
+  }
+
+  if (code !== 0) {
+    throw new Error(
+      `Command failed with exit code ${code}: ${decoder.decode(stderr)}`
+    );
+  }
+
+  return decoder.decode(stdout).trim();
 }
```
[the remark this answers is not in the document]
akhileshthite: Yes, this will be resolved in a separate linting PR pointing to this one.

src/deps.ts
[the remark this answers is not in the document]
akhileshthite: 1.  The JSR esbuild is not maintained, that’s the reason i’m using latest version of deno land
https://jsr.io/@build/esbuild
2. As I mentioned in other comments, this will be resolved in the linting PR.

deno.json
akhileshthite: https://github.com/okTurtles/chel/pull/46/commits/54a9fa20b38171178935f7d056540a002ac32a22

src/version.ts
```diff
@@ -1,6 +1,3 @@
-'use strict'
-
 export function version () {
-  // @ts-ignore: "cannot find name 'process'"
-  console.log(process.env.VERSION)
+  console.log(globalThis.VERSION);
```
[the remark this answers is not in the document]
akhileshthite: https://github.com/okTurtles/chel/pull/46/commits/df8066f42e746365022335609261a73ba9797e4b

scripts/build.ts
```diff
@@ -10,12 +10,23 @@ const options = {
   entryPoints: ['./src/main.ts'],
   bundle: true,
   define: {
-    'process.env.VERSION': `'${version}'`
+    "globalThis.VERSION": JSON.stringify(version),
```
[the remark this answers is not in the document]
akhileshthite: https://github.com/okTurtles/chel/pull/46/commits/df8066f42e746365022335609261a73ba9797e4b

corrideat:
Great! It looks like all of my comments were addressed, but please note the whitespace issue with `build.d.ts`.

akhileshthite:
> Great! It looks like all of my comments were addressed, but please note the whitespace issue with `build.d.ts`.

Thanks!
Yes, all the styling issues — including the whitespace — will be resolved in the linting PR.

corrideat:
> Yes, all the styling issues — including the whitespace — will be resolved in the linting PR.

Sure, but there's probably little reason to leave known linting issues on new code on this PR. If you make a new commit to this branch, it might be a good idea to also fix the whitespace.

And, you will need at least one more commit to this PR. The changes look fine to me, but CI tests are failing because CI is still using Deno 1. `deno-version` in `.github/workflows/ci.yml` needs to be updated to a Deno 2 version.

There's also another issue: `vendor/binary-install/index.js` has been removed, and, because of this, `npm install` no longer works.

If you get tests working and fix the `npm install` issue, I'd say this PR is ready to merge.

akhileshthite:
> There's also another issue: `vendor/binary-install/index.js` has been removed, and, because of this, `npm install` no longer works.
> 
> If you get tests working and fix the `npm install` issue, I'd say this PR is ready to merge.

Fixed the tests, most of the CI errors are addressed in #47 and will go away when we merge.

Regarding the `vendor/binary-install/index.js` I think with the `deno v2` it removed when I ran builds.

Do you want me to add a script `binary-install` for this? Like in `./scripts/` as `binary-install` which will be included in deno.json `tasks`

src/database-sqlite.ts
akhileshthite, line 1: @corrideat 
Fixed SQLite backend initialization by adapting to the `jsr:@db/sqlite@0.12.0 API`, replacing execute/prepareQuery with run/prepare and manual statement handling to resolve test failures. https://github.com/okTurtles/chel/pull/46/commits/1faad211879f8ed93f03ca073ae6b7578d5c47d0

`deno test --allow-read --allow-write=test --allow-env --allow-ffi --allow-net --no-check`:
```
Passing:
running 1 test from ./test/getBackend.test.ts
Tests for getBackend() ...
  it should fail given a non-existing folder if option `create` was not `true` ... ok (1ms)
  FS - should fail given a file instead of a folder ... ok (0ms)
  SQLite - should fail given a folder instead of a file ... ok (2ms)
  FS - it should open an existing folder as `fs` by default ... ok (0ms)
  FS - it should create the folder if missing and option `create` is `true` ... ok (0ms)
  SQLite - it should open an existing .db file as `sqlite` by default ... ok (6ms)
  SQLite - it should create the .db file if missing and option `create` is `true` ... ok (2ms)
Tests for getBackend() ... ok (13ms)
running 1 test from ./test/hash.test.ts
Hash command tests ...
  it should create a raw CID for a plain text file ... ok (1ms)
  it should create a raw CID for a json file ... ok (1ms)
Hash command tests ... ok (2ms)
running 1 test from ./test/signature.test.ts
Verify signature tests ...
  it should validate a valid signature ... ok (17ms)
  it should validate a valid signature (using a matching external key descriptor file) ... ok (8ms)
  it should not validate a valid signature when using the wrong key file ... ok (1ms)
  it should not validate a valid signature if the public key isn't enumerated ... ok (5ms)
  it should not validate a valid signature missing its key ID ... ok (1ms)
  it should not validate a valid signature with a wrong key ID ... ok (1ms)
  it should not validate a manifest file that's been tampered with ... ok (9ms)
  it should not validate a tampered with bundle contract ... ok (10ms)
  it should not validate a tampered with slim contract ... ok (32ms)
Verify signature tests ... ok (87ms)

ok | 3 passed (18 steps) | 0 failed (605ms)
```

src/types/build.d.ts
```diff
@@ -1,4 +1,4 @@
 declare namespace __build__ {
     const VERSION: string
-  }
+}
```
corrideat, line 4: Extra whitespace here

.github/workflows/ci.yml
```diff
@@ -14,5 +14,8 @@ jobs:
     - uses: actions/checkout@v3
     - uses: denoland/setup-deno@v1
       with:
-        deno-version: v1.40.4
-    - run: deno task test --no-check
+        deno-version: v2.4.0
```
taoeffect, line 17: Should `denoland/setup-deno@v1` on line 15 be updated as well? What's the latest version?

corrideat: The latest version is 2.0.3. I'd also say that this is an opportunity to use commit IDs (immutable) instead of tags (mutable).

akhileshthite: Yeah, it's v2 https://github.com/denoland/setup-deno

akhileshthite: > The latest version is 2.0.3. I'd also say that this is an opportunity to use commit IDs (immutable) instead of tags (mutable).

Like `denoland/setup-deno@e95548e56dfa95d4e1a28d6f422fafe75c4c26fb`?

taoeffect: I think that is the syntax, you can double-check with LLMs. I would also add a comment in addition to that (on the line above) specifying what version that represents

src/deps.ts
```diff
@@ -1,21 +1,17 @@
...
+export * as flags from 'jsr:@std/flags@0.224.0';
+export * as colors from 'jsr:@std/fmt@1.0.8/colors';
+export * as fs from 'jsr:@std/fs@1.0.19';
+export * as path from 'jsr:@std/path@1.1.1';
+export * as streams from 'jsr:@std/streams@1.0.10';
+export * as util from 'jsr:@std/io@0.225.2';
+export { copy, readAll } from 'jsr:@std/io@0.225.2';
+export * as sqlite from "jsr:@db/sqlite@0.12.0";
+export { type Database as SQLiteDB } from "jsr:@db/sqlite@0.12.0";
+export * as esbuild from "https://deno.land/x/esbuild@v0.25.6/mod.js";
```
taoeffect, line 12: This is the last deno.land dependency. Could you grab it from npm instead? (remembering to specify precise version)

deno.json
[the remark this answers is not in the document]
taoeffect: Your other PR #47 can be merged into this one right? Then you can remove `--no-check` here too, right?

See my request there to do so: https://github.com/okTurtles/chel/pull/47#pullrequestreview-3014268840

src/database-sqlite.ts
```diff
@@ -1,14 +1,14 @@
...
 let db: SQLiteDB
 let dbPath: string
-let iterKeysStatement: PreparedQuery<[string]>
-let readStatement: PreparedQuery<[string]>
-let writeOnceStatement: PreparedQuery
-let writeStatement: PreparedQuery
+let iterKeysStatement: sqlite.Statement;
+let readStatement: sqlite.Statement;
+let writeOnceStatement: sqlite.Statement;
+let writeStatement: sqlite.Statement;
```
taoeffect, line 11: All of these semi-colons are `standard` violations

src/deps.ts
```diff
@@ -1,21 +1,17 @@
...
+export { assert, assertEquals, assertRejects, assertThrows } from 'jsr:@std/assert@1.0.13';
+export * as base64 from 'jsr:@std/encoding@1.0.10/base64';
+export * as flags from 'jsr:@std/flags@0.224.0';
+export * as colors from 'jsr:@std/fmt@1.0.8/colors';
+export * as fs from 'jsr:@std/fs@1.0.19';
+export * as path from 'jsr:@std/path@1.1.1';
+export * as streams from 'jsr:@std/streams@1.0.10';
+export * as util from 'jsr:@std/io@0.225.2';
+export { copy, readAll } from 'jsr:@std/io@0.225.2';
+export * as sqlite from "jsr:@db/sqlite@0.12.0";
```
taoeffect, line 10: Is this the same as https://www.npmjs.com/package/sqlite ?

If so, I'm pretty sure there's a reason why we switched off it to better-sqlite. And then in this PR @corrideat is using `node:sqlite` because of some issue with Deno: https://github.com/okTurtles/group-income/pull/2759/files#diff-41886e760ca1da4471f8c675c59fa693366ef97a98ebd13dd90f647d1f7de729R6

@corrideat can you confirm what should be done here?

taoeffect (changes requested):
Nice work @akhileshthite, looks like we're almost ready to merge.

I recommend merging the changes from PR #47 into there now as opposed to later. Removing the `--no-check` so that we can verify the type checks pass before merging. And using the correct sqlite package. See also other comments:

src/deps.ts
taoeffect: @corrideat Except it is a concern... since we're planning on moving the entire backend into here. So which should we use?

src/deps.ts
akhileshthite: https://github.com/okTurtles/chel/pull/46/commits/41c58ec788acef6e841c9153bbe246d11947903f

.github/workflows/ci.yml
akhileshthite: https://github.com/okTurtles/chel/pull/46/commits/4199b2739d110b06c59c21467d5f5863fd408e4c

akhileshthite:
Fixed these errors:

```
TS2554 [ERROR]: Expected 0 arguments, but got 1.
    db.close(true)
             ~~~~
    at database-sqlite.ts:28:14

TS2558 [ERROR]: Expected 0 type arguments, but got 1.
  iterKeysStatement = db.prepare<[string]>('SELECT key FROM Data')
                                 ~~~~~~~~
    at database-sqlite.ts:37:34

TS2558 [ERROR]: Expected 0 type arguments, but got 1.
  readStatement = db.prepare<[string]>('SELECT value FROM Data WHERE key = ?')
                             ~~~~~~~~
    at database-sqlite.ts:38:30

TS2339 [ERROR]: Property 'query' does not exist on type 'Database'.
  return db.query<[number]>('SELECT COUNT(*) FROM Data')[0][0]
            ~~~~~
    at database-sqlite.ts:46:13

TS2339 [ERROR]: Property 'first' does not exist on type 'Statement'.
  const maybeRow = readStatement.first([key])
                                 ~~~~~
    at database-sqlite.ts:52:34

TS2339 [ERROR]: Property 'execute' does not exist on type 'Statement'.
  writeStatement.execute([key, value])
                 ~~~~~~~
    at database-sqlite.ts:65:18

TS2339 [ERROR]: Property 'execute' does not exist on type 'Statement'.
  writeOnceStatement.execute([key, value])
                     ~~~~~~~
    at database-sqlite.ts:71:22

TS2305 [ERROR]: Module '"file:///Users/akhileshthite/Documents/chel/src/deps.ts"' has no exported member 'writeAll'.
import { flags, writeAll } from './deps.ts'
                ~~~~~~~~
    at get.ts:11:17

TS2304 [ERROR]: Cannot find name '__build__'.
  console.log(__build__.VERSION)
              ~~~~~~~~~
    at version.ts:2:15

Found 9 errors.

error: Type checking failed.
```

build/main.js
```diff
@@ -118,7 +117,7 @@ async function initStorage2(options = {}) {
     if (filepath === dbPath) {
       return;
     }
-    db.close(true);
+    db.close();
```
akhileshthite, line 120: The close() method no longer takes any arguments.

src/database-sqlite.ts
```diff
@@ -34,22 +34,22 @@ export async function initStorage (options: Record<string, unknown> = {}): Promi
   if (!options.internal) {
     console.log('Connected to the %s SQLite database.', filepath)
   }
-  iterKeysStatement = db.prepare<[string]>('SELECT key FROM Data')
-  readStatement = db.prepare<[string]>('SELECT value FROM Data WHERE key = ?')
+  iterKeysStatement = db.prepare('SELECT key FROM Data')
+  readStatement = db.prepare('SELECT value FROM Data WHERE key = ?')
```
akhileshthite: Type inference now handles this automatically.

src/database-sqlite.ts
```diff
@@ -34,22 +34,22 @@ export async function initStorage (options: Record<string, unknown> = {}): Promi
...
 export function count (): number {
-  return db.query<[number]>('SELECT COUNT(*) FROM Data')[0][0]
+  return db.prepare('SELECT COUNT(*) FROM Data').all()[0][0]
 }
 
 // deno-lint-ignore require-await
 export async function readData (key: string): Promise<Uint8Array | string | void> {
   // For some reason `[null]` is returned when the value is an empty Uint8Array.
-  const maybeRow = readStatement.first([key])
+  const maybeRow = readStatement.all([key])[0]
```
akhileshthite, line 52: Statement.first() is removed; used .all() and accessed the first element manually.

src/database-sqlite.ts
```diff
@@ -62,11 +62,11 @@ export async function* iterKeys (): AsyncGenerator<string> {
...
   checkKey(key)
-  writeStatement.execute([key, value])
+  writeStatement.run([key, value])
 }
 
 // deno-lint-ignore require-await
 export async function writeDataOnce (key: string, value: Uint8Array | string): Promise<void> {
   checkKey(key)
-  writeOnceStatement.execute([key, value])
+  writeOnceStatement.run([key, value])
```
akhileshthite, line 71: .execute() is no longer supported; used .run() instead like we did previously.

taoeffect (approved):
Approved! 🎉

src/database-sqlite.ts
corrideat: Is this guaranteed to be defined?

akhileshthite: Right, it's not guaranteed to be defined. If the query doesn't find a row matching the key, readStatement.all([key]) will return an empty array.

We should do `const maybeRow = readStatement.all([key])[0] ?? undefined`

taoeffect: Hmm, if it returns an empty array, is it necessary to do that?

```
> [][0] ?? undefined
undefined
> [][0]
undefined
> 
```

akhileshthite: https://github.com/okTurtles/chel/pull/46/commits/f6dde63cb1ae5875b056c3ddf067edd8049822da

corrideat: Sorry if it wasn't clear. https://github.com/okTurtles/chel/commit/f6dde63cb1ae5875b056c3ddf067edd8049822da should be reverted because what it replaced has the same effect. My question was about whether `readStatement.all([key])[0]` would always succeed (e.g., maybe `readStatement.all([key])` was `undefined` or `null`).

akhileshthite: Should we keep as is?
`const maybeRow = readStatement.all([key])[0]`

akhileshthite: Reverted https://github.com/okTurtles/chel/pull/46/commits/3cedbd2bcd313d6852344306f7829d61d592504e

.github/workflows/ci.yml
```diff
@@ -12,7 +12,11 @@ jobs:
...
+    # denoland/setup-deno v2.3.0
+    - uses: denoland/setup-deno@e95548e56dfa95d4e1a28d6f422fafe75c4c26fb
       with:
-        deno-version: v1.40.4
-    - run: deno task test --no-check
+        deno-version: v2.4.0
+    - run: deno cache --lock=deno.lock src/deps.ts
+    
+    - name: Run tests
+      run: deno test --allow-read --allow-write --allow-env --allow-ffi --allow-net
```
corrideat, line 22: Missing line feed

corrideat (approved):
I think this is ready to be approved, but please add a line feed to the CI YAML file.

bin/chel
```diff
@@ -1 +1 @@
-placeholder, will be replaced by real binary. DO NOT DELET THIS FILE!
+placeholder, will be replaced by real binary. DO NOT DELETE THIS FILE!
```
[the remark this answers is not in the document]
taoeffect: Hmm... Nice catch! I would fix that just to be safe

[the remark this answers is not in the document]
corrideat: If installing the binary sets the right permission, I think this is perhaps better as it'd prevent accidental execution.

PR #47, opened 2025-07-10
fix: CLI bugs and improve type safety across utils and chel src

akhileshthite:
Question regarding this:
Can we guarantee `kf` is always a `string[]`? If so, I’ll type `kf: string` directly. Otherwise, I’ll keep `unknown` and convert it with `String()`.

```sh
TS2345 [ERROR]: Argument of type '(kf: number | string) => Promise<string>' is not assignable to parameter of type '(value: unknown, index: number, array: unknown[]) => Promise<string>'.
  Types of parameters 'kf' and 'value' are incompatible.
    Type 'unknown' is not assignable to type 'string | number'.
        async (kf: number | string) => {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    at /chel/src/manifest.ts:34:9

error: Type checking failed.
```


akhileshthite:
> The type depends on what `flags` gives out, and I think we should mostly rely on the type it gives out. If the type has changed from `string | number` to `unknown`, then what makes the most sense is to use `unknown` and check whether it's a string. But, first I'd double check where `unknown` is coming from and why by checking the source (or types) of `flags`.

I checked the `flags.parse()` types, and you're right — with `{ collect: ['key'] }`, the returned type is `unknown[]`. I've updated the code to treat `kf` as `unknown` and safely coerce it using `String(kf)`.

https://github.com/okTurtles/chel/pull/47/commits/e419077b0caa6e115da34051529a1be0f527e4e1

src/eventsAfter.ts
```diff
@@ -28,7 +28,7 @@ export async function eventsAfter (args: string[]): Promise<void> {
     }
     console.log(JSON.stringify(messages, null, 2))
   } catch (error) {
-    exit(error.message)
+    exit((error as Error).message)
```
corrideat, line 31: Instead of all of these `(x as Error)` casts, it'd make more sense to modify `exit` to take `unknown`, and do as follows:

1. `if (typeof x === 'string') msg = x; else if (x instanceof Error) msg = x.message; else msg = String(x)`

src/database-fs.ts
```diff
@@ -60,6 +60,6 @@ export async function writeDataOnce(key: string, value: Uint8Array | string): Pr
       await Deno.writeFile(path.join(dataFolder, key), value, options)
     }
   } catch (err) {
-    if (err.name !== 'AlreadyExists') throw err
+    if ((err as Error).name !== 'AlreadyExists') throw err
```
corrideat, line 63: Since the code is uglier anyhow due to the cast, `err instanceof Error && err.name !== 'AlreadyExists'` is a more robust solution.

src/utils.ts
```diff
@@ -78,7 +84,7 @@ export async function getBackend (src: string, { type, create } = { type: '', cr
   try {
     await backend.initStorage(initOptions)
   } catch (error) {
-    throw new Error(`could not init '${from}' storage backend at "${src}": ${error.message}`)
+    throw new Error(`could not init '${from}' storage backend at "${src}": ${error}`)
```
corrideat, line 87: This is one of the places where, to keep the old behaviour, casting is needed, since `String(error)` is different from `error.message` (typically it'll be `{error.name}: {error.message}\n{error.stack}`)

corrideat: I'd wait on @taoeffect to say what should be done here re. the message. Some options are removing the last part (leaving it at `could not init '${from}' storage backend at "${src}"`), keeping it as it was or leaving it as it currently is.

However, I'd request one change regardless of what's done with the string, which is to have `throw new Error(..., { cause: e })`. This embeds the previous error into the new one, which can be useful for debugging.

src/manifest.ts
```diff
@@ -71,7 +74,7 @@ export async function manifest (args: string[]) {
     console.log(manifest)
   } else {
     const outFile = parsedArgs.out || outFilepath
-    Deno.writeTextFileSync(outFile, manifest)
+    Deno.writeTextFileSync(String(outFile), manifest)
```
taoeffect, line 77: I'm not a big fan of this `String(outFile)` style casting because it requires doing it each time the variable is used.

Look at lines 58 and 59 for an example.

Is it possible to do this instead?

```js
const outFile: string = parsedArgs.out || outFilepath
```

Admittedly, I am only really familiar with Flowtype, not TypeScript, but whatever the TypeScript equivalent is.

Same question for lines 58 - 59 above for `slim`. Could something along these lines be done instead?

```js
const slim: string | undefined = parsedArgs.slim || parsedArgs.s
```

src/utils.ts
taoeffect: I don't understand why this line was changed.

I'm not opposed to adding a `{ cause: e }`, but would like to see this line reverted. If it has something to do with `e.message` not working with typing, then please use whatever cast makes TypeScript happy, e.g. `e.message as string` or `String(e.message)` or `e.message || ''` (I'm not super familiar with TypeScript yet)

src/utils.ts
```diff
@@ -44,9 +44,15 @@ export function createCID (data: string | Uint8Array, multicode = multicodes.RAW
...
 }
 
-export function exit (message: string, internal = false): never {
-  if (internal) throw new Error(message)
-  console.error('[chel]', colors.red('Error:'), message)
+export function exit(x: unknown, internal = false): never {
+  let msg: string
+  if (typeof x === 'string') msg = x
+  else if (x instanceof Error) msg = x.message
+  else msg = String(x)
```
taoeffect, line 51: This can be shortened to 1 line:

```js
const msg = x instanceof Error ? x.message : String(x)
```

taoeffect (changes requested):
Thanks @akhileshthite!

I left a review. Could you consider that after applying the changes for the review, that these changes be merged into #46 as this is a "naked PR", unattached to any issue, and it's also rather short anyway.

src/utils.ts
corrideat: > (I'm not super familiar with TypeScript yet.

This is not so much a of TypeScript issue, except that TypeScript makes it evident.

JavaScript doesn't impose any specific requirements on what can be thrown, as you know. So, `throw 1`, `throw Error('test')`, `throw undefined`, `throw 'Error'`, `throw {}`, and so on, are all valid statements. The result of this is that the only type that can be safely used is `unknown` 
(in Flow, this would be `mixed`).

Because of the above, `${error.message}` doesn't safely work, because it's not a given that `.message` will be a valid property. There are basically two ways to address this.

* `${(error as Error).message}`: this will cast `error` to `Error`, making the TypeScript error go away (although the code is still technically incorrect, as it can't be guaranteed that this assertion is correct.
* `${error instanceof Error ? error.message : error}`: This is slightly better as it'll almost never go wrong or itself result in an exception.

src/manifest.ts
akhileshthite: parsedArgs values are typed as `unknown` by default, so direct assignment to `string` fails type checking; we use type assertions to tell TypeScript they're strings.
https://github.com/okTurtles/chel/pull/47/commits/ceae3abd58e0aa3baead94ed03c58b80f7c6cebb

taoeffect (approved):
Nice!

PR #49, opened 2025-07-14
style: Implement Standard Linting with ts-standard and TypeScript Integration

akhileshthite:
Closes https://github.com/okTurtles/chel/issues/45

PR #51, opened 2025-07-14
ci: run ts-standard linter in GitHub Actions

### LINKED ISSUE #50: Run the standard linter on CI
### Problem

The `ts-standard` linter doesn't run on Github Actions.

This can make us miss bugs.

### Solution

Please run it on Github actions either by making sure that `deno test` runs it, or by updating the workflow file.

akhileshthite:
Closes #50

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,9 @@ jobs:
       with:
         deno-version: v2.4.0
     - run: deno cache --lock=deno.lock src/deps.ts
+
+    - name: Run ts-standard linter
+      run: npx ts-standard ./src ./scripts
     
     - name: Run tests
       run: deno test --allow-read --allow-write --allow-env --allow-ffi --allow-net


scripts/compile.ts
```diff
@@ -2,23 +2,23 @@
...
     const dir = `./dist/tmp/${arch}`
     const bin = arch.includes('windows') ? 'chel.exe' : 'chel'
     // note: could also use https://examples.deno.land/temporary-files
     await $(`mkdir -vp ${dir}`)
     await $(`deno compile --allow-read=./ --allow-write=./  --allow-net -o ${dir}/${bin} --target ${arch} ./build/main.js`)
-    await $(`tar -C ./dist/tmp -czvf ./dist/chel-v${version}-${arch}.tar.gz ${arch}`)
+    await $(`tar -C ./dist/tmp -czvf ./dist/chel-v${version as string}-${arch}.tar.gz ${arch}`)
   }
-  await $(`sha256sum dist/chel-v${version}-*`)
+  await $(`sha256sum dist/chel-v${version as string}-*`)
```
taoeffect, line 21: Instead of adding `as string` multiple times, is it possible to add the type info on line 9?

src/database-fs.ts
```diff
@@ -20,26 +20,33 @@ export async function clear (): Promise<void> {
 
 export async function count (): Promise<number> {
   let n = 0
-  for await (const _entry of Deno.readDir(dataFolder)) {
-    n++
+  for await (const entry of Deno.readDir(dataFolder)) {
+    // Explicit check to satisfy strict-boolean-expressions lint rule
+    if (entry.isFile === true) {
```
taoeffect, line 25: I am not a fan of this `strict-boolean-expressions` rule. It is causing ugly modifications throughout the code.

Could you keep the check as `if (entry.isFile) {` and place `strict-boolean-expressions` on the ignore list?

src/database-fs.ts
```diff
@@ -20,26 +20,33 @@ export async function clear (): Promise<void> {
...
+    }
   }
   return n
 }
 
 export async function * iterKeys (): AsyncGenerator<string> {
   for await (const entry of Deno.readDir(dataFolder)) {
     // Skip subfolders and symlinks.
-    if (entry.isFile) {
+    if (entry.isFile === true) {
```
taoeffect, line 35: Then this can be reverted

src/database-fs.ts
```diff
@@ -20,26 +20,33 @@ export async function clear (): Promise<void> {
...
-export async function readData (key: string): Promise<Uint8Array | string | void> {
-  // Necessary here to thwart path traversal attacks.
+export async function readData (key: string): Promise<Uint8Array | string | undefined> {
   checkKey(key)
-  return await Deno.readFile(path.join(dataFolder, key))
-    .catch(_err => undefined)
+  try {
+    // assign to a variable before returning to satisfy no-return-await
+    const file = await Deno.readFile(path.join(dataFolder, key))
+    return file
```
taoeffect, line 46: I think this `no-return-await` can be avoided also by reverting the logic to as it was and removing the `async` / `await`

src/deploy.ts
```diff
@@ -9,17 +9,21 @@ import { upload } from './upload.ts'
...
   const [urlOrDirOrSqliteFile, ...manifests] = args
   if (manifests.length === 0) throw new Error('missing url or manifests!')
   const toUpload = []
   for (const manifestPath of manifests) {
-    const json = JSON.parse(Deno.readTextFileSync(manifestPath))
-    const body = JSON.parse(json.body)
+    interface ContractBody {
+      contract: { file: string }
+      contractSlim?: { file: string }
+    }
```
taoeffect, line 20: A few questions here:

- Can this definition be moved above and outside of this function?
- Is this really necessary? If so, I'm a bit skeptical of it as it's not really enforced upon parsing. [Something like Zod](https://zod.dev/) would make more sense as that would actually enforce runtime checks if you wanted them

src/deploy.ts
```diff
@@ -9,17 +9,21 @@ import { upload } from './upload.ts'
...
+    const json = JSON.parse(Deno.readTextFileSync(manifestPath)) as { body: string }
+    const body = JSON.parse(json.body) as ContractBody
     const dirname = path.dirname(manifestPath)
-    toUpload.push(CONTRACT_TEXT_PREFIX + path.join(dirname, body.contract.file))
-    if (body.contractSlim) {
-      toUpload.push(CONTRACT_TEXT_PREFIX + path.join(dirname, body.contractSlim.file))
+    toUpload.push(CONTRACT_TEXT_PREFIX + String(path.join(dirname, body.contract.file)))
+    if (body.contractSlim !== undefined && body.contractSlim !== null) {
+      toUpload.push(CONTRACT_TEXT_PREFIX + String(path.join(dirname, body.contractSlim.file)))
```
taoeffect, line 26: Could these changes be reverted?

- The `!== undefined` and `!== null` can be reverted to as it was by ignoring `strict-boolean-expressions`. The previous check already did that in a cleaner way. We don't want to uglify our code for zero reason.
- These `String` coercions are also obnoxious, what rule is causing you to add them? According to [the `path.join` docs](https://nodejs.org/docs/latest/api/path.html#pathjoinpaths), that function returns a `string`, guaranteed, so why are we calling `String` on it? Whatever rule is making you do this, please see if you can ignore it too

src/eventsAfter.ts
```diff
@@ -34,14 +34,14 @@ export async function eventsAfter (args: string[]): Promise<void> {
 
 async function getMessage (hash: string): Promise<ReturnType<typeof JSON.parse>> {
   const value = await readString(hash)
-  if (!value) throw new Error(`no entry for ${hash}!`)
+  if (value === undefined || value === null) throw new Error(`no entry for ${hash}!`)
```
taoeffect, line 37: Same request for reverting these (and all other ones, I'll stop repeating myself, but please check the rest of the codebase for this)

src/eventsAfter.ts
```diff
@@ -68,17 +68,17 @@ async function getRemoteMessagesSince (src: string, contractID: string, sinceHei
   const response = await fetch(`${src}/eventsAfter/${contractID}/${sinceHeight}`)
   if (!response.ok) {
     // The response body may contain some useful error info if we got a Boom error response.
-    const bodyText = await response.text().catch(_ => '') || ''
+    const bodyText = await response.text().catch(_ => '') !== undefined ? await response.text().catch(_ => '') : ''
```
taoeffect, line 71: This is really weird, can you revert this is as well?

src/eventsAfter.ts
```diff
@@ -68,17 +68,17 @@ async function getRemoteMessagesSince (src: string, contractID: string, sinceHei
...
     throw new Error(`failed network request to ${src}: ${response.status} - ${response.statusText} - '${bodyText}'`)
   }
   const b64messages: string[] = await response.json()
   if (b64messages.length > limit) {
     b64messages.length = limit
   }
-  return b64messages.map(b64str => JSON.parse(new TextDecoder().decode(base64.decodeBase64(b64str))))
+  return b64messages.map(b64str => b64str !== '' ? JSON.parse(new TextDecoder().decode(base64.decodeBase64(b64str))) : null)
```
taoeffect, line 78: This change doesn't make sense to me. Can revert?

src/help.ts
```diff
@@ -18,7 +18,11 @@ export function help (args?: string[]) {
       chel hash <file>
       chel migrate --from <backend> --to <backend> --out <dir-or-sqlitedb> <dir-or-sqlitedb>
     `)
-  } else if (helpDict[args[0]]) {
+  } else if (
+    typeof args[0] === 'string' &&
+    args[0].trim() !== '' &&
+    Object.prototype.hasOwnProperty.call(helpDict, args[0])
```
taoeffect, line 24: Don't understand why this was uglified

src/keygen.ts
```diff
@@ -18,8 +17,12 @@ export const keygen = async (args: string[]) => {
...
 
   const idx = keyId(key).slice(-12)
-  const outFile = parsedArgs.out || `${EDWARDS25519SHA512BATCH}-${idx}.json`
-  const pubOutFile = parsedArgs.pubout || `${EDWARDS25519SHA512BATCH}-${idx}.pub.json`
+  const outFile: string = typeof parsedArgs.out === 'string' && parsedArgs.out.trim() !== ''
+    ? parsedArgs.out
+    : `${String(EDWARDS25519SHA512BATCH)}-${String(idx)}.json`
+  const pubOutFile: string = typeof parsedArgs.pubout === 'string' && parsedArgs.pubout.trim() !== ''
+    ? parsedArgs.pubout
+    : `${String(EDWARDS25519SHA512BATCH)}-${String(idx)}.pub.json`
```
taoeffect, line 25: What can be done to de-uglify this and restore it to how it was?

src/main.ts
```diff
@@ -15,12 +15,21 @@ import * as commands from './commands.ts'
...
+  typeof command === 'string' &&
+  Object.prototype.hasOwnProperty.call(commands, command)
+) {
+  const cmdFn = commands[command as keyof typeof commands]
+  if (typeof cmdFn === 'function') {
+    await cmdFn(rest)
+  } else {
+    console.error(`Command exists but is not callable: ${String(command)}`)
+    Deno.exit(1)
+  }
```
taoeffect, line 30: Hmm.... can this be simplified anyhow?

src/manifest.ts
```diff
@@ -5,55 +5,70 @@
...
+  } else if (typeof parsedArgs.s === 'string') {
+    slim = parsedArgs.s
+  }
+  const outFile: string =
+  typeof parsedArgs.out === 'string'
+    ? parsedArgs.out
+    : path.join(contractDir, `${String(contractFileName)}.${String(version)}.manifest.json`)
+  if (typeof keyFileRaw !== 'string' || typeof contractFileRaw !== 'string') {
+    exit('Missing or invalid key or contract file')
+  }
```
taoeffect, line 39: Can this be simplified or restored to how it was?

src/manifest.ts
```diff
@@ -5,55 +5,70 @@
...
-            exit(`Invalid key file reference: ${String(kf)}`)
-          }
-          const descriptor = await readJsonFile(String(kf))
-          const key = deserializeKey(descriptor.pubkey)
-          if (key.type !== EDWARDS25519SHA512BATCH) {
-            exit(`Invalid key type ${key.type}; only ${EDWARDS25519SHA512BATCH} keys are supported.`)
-          }
-          return serializeKey(key, false)
+    [serializeKey(signingKey, false)].concat(...await Promise.all(
+      (additionalKeys ?? []).map(async (kf) => {
```
taoeffect, line 49: Can this be made cleaner?

Either reverted or something like:

```js
const additionalKeys = parsedArgs.key || [] as Array<string | number>
```

src/manifest.ts
```diff
@@ -5,55 +5,70 @@
...
+        if (typeof kf !== 'string' && typeof kf !== 'number') {
+          exit(`Invalid key file reference: ${String(kf)}`)
+        }
+        const descriptor = await readJsonFile(String(kf)) as { pubkey: string }
+        const key = deserializeKey(descriptor.pubkey)
+        if (key.type !== EDWARDS25519SHA512BATCH) {
+          exit(`Invalid key type ${String(key.type)}; only ${String(EDWARDS25519SHA512BATCH)} keys are supported.`)
```
taoeffect, line 56: Why is `String` needed in front of these? Is that type info not there?

src/manifest.ts
```diff
@@ -5,55 +5,70 @@
...
+        return serializeKey(key, false)
+      })
+    ))
   ))
   const body: { [key: string]: unknown } = {
     name,
     version,
     contract: {
-      hash: await hash([contractFile as string], multicodes.SHELTER_CONTRACT_TEXT, true),
+      hash: await hash([contractFile], multicodes.SHELTER_CONTRACT_TEXT, true),
```
taoeffect, line 66: Nice, a simplification!

src/migrate.ts
```diff
@@ -46,5 +46,5 @@ export async function migrate (args: string[]) {
       console.log(`[chel] Migrating... ${Math.round(numVisitedKeys / (numKeys / 10))}0% done`)
     }
   }
-  numKeys && console.log(`[chel] ${colors.green('Migrated:')} ${numKeys} entries`)
+  if (numKeys > 0) console.log(`[chel] ${(colors.green as (text: string) => string)('Migrated:')} ${numKeys} entries`)
```
taoeffect, line 49: This `(colors.green as (text: string) => string` thing is insanely ugly, can you get rid of it somehow?

src/upload.ts
```diff
@@ -40,7 +40,7 @@ export async function upload (args: string[], internal = false) {
     if (!internal) {
       console.log(colors.green('uploaded:'), destination)
     } else {
-      console.log(colors.green(`${path.relative('.', filepath)}:`), destination)
+      console.log(colors.green(`${String(path.relative('.', filepath))}:`), destination)
```
taoeffect, line 43: `path.relative` should be typed to return a `string`, so again like with `path.join` I don't get why this change was made, and if there's a way to get rid of it, it would be appreciated

src/upload.ts
```diff
@@ -71,8 +71,8 @@ async function uploadEntryToDir ([cid, buffer]: Entry, dir: string): Promise<str
 async function uploadEntryToSQLite ([cid, buffer]: Entry, sqlitedb: string): Promise<string> {
   await revokeNet()
   const { initStorage, writeData } = await import('./database-sqlite.ts')
-  initStorage({ dirname: path.dirname(sqlitedb), filename: path.basename(sqlitedb) })
-  writeData(cid, buffer)
+  await initStorage({ dirname: path.dirname(sqlitedb), filename: path.basename(sqlitedb) })
+  await writeData(cid, buffer)
```
taoeffect, line 75: This looks like a bugfix actually

src/verifySignature.ts
```diff
@@ -1,78 +1,115 @@
...
-  if (!manifest.signature) {
+  if (typeof manifest.signature !== 'object' || manifest.signature === null) {
     exit('Invalid manifest: missing signature', internal)
   }
-  if (!manifest.signature.keyId) {
+  if (typeof manifest.signature.keyId !== 'string' || manifest.signature.keyId === '') {
     exit('Invalid manifest: missing signature key ID', internal)
   }
-  if (!manifest.signature.value) {
+  if (typeof manifest.signature.value !== 'string' || manifest.signature.value === '') {
```
taoeffect, line 64: Can these be reverted?

src/verifySignature.ts
```diff
@@ -1,78 +1,115 @@
...
+  if (typeof body.contract.hash !== 'string' || computedHash !== body.contract.hash) {
+    exit(`Invalid contract file hash. Expected ${String(body.contract.hash)} but got ${computedHash}`, internal)
   }
 
-  if (body.contractSlim) {
+  if (typeof body.contractSlim === 'object' && body.contractSlim !== null) {
     const computedHash = await hash([path.join(parsedFilepath.dir, body.contractSlim.file)], multicodes.SHELTER_CONTRACT_TEXT, true)
     if (computedHash !== body.contractSlim.hash) {
-      exit(`Invalid slim contract file hash. Expected ${body.contractSlim.hash} but got ${computedHash}`, internal)
+      exit(`Invalid slim contract file hash. Expected ${String(body.contractSlim.hash)} but got ${computedHash}`, internal)
```
taoeffect, line 112: Same request for reversion..

taoeffect (changes requested):
Nice work @akhileshthite! It seems like TypeScript is giving us some serious headaches 🤕

Review is ready, and also copying my question here from Slack in case you missed it, but feel free to reply either here or on Slack:

>  could you help me understand why you've had to make so many type changes in this PR when your [other PR](https://github.com/okTurtles/chel/pull/49) also used ts-standard and made type changes?

src/utils.ts
```diff
@@ -122,9 +122,8 @@ export function isURL (arg: string): boolean {
 }
 
 export function isValidKey (key: string): boolean {
-  // Disallow unprintable characters, slashes, and TAB.
-  // deno-lint-ignore no-control-regex
-  return !/[\x00-\x1f\x7f\t\\/]/.test(key)
+  // Accept only printable ASCII except slash/backslash
+  return /^[\x20-\x7E]*$/.test(key) && !/[\\/]/.test(key)
```
taoeffect, line 126: Why was this regex modified?

src/database-fs.ts
```diff
@@ -20,26 +20,28 @@ export async function clear (): Promise<void> {
...
     if (entry.isFile) {
       yield entry.name
     }
   }
 }
 
-export async function readData (key: string): Promise<Uint8Array | string | void> {
-  // Necessary here to thwart path traversal attacks.
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export function readData (key: string) {
```
taoeffect, line 42: The return type here should be kept — having return types defined is good practice

src/database-fs.ts
```diff
@@ -20,26 +20,28 @@ export async function clear (): Promise<void> {
...
+    if (entry.isFile) {
+      n++
+    }
   }
   return n
 }
 
 export async function * iterKeys (): AsyncGenerator<string> {
   for await (const entry of Deno.readDir(dataFolder)) {
-    // Skip subfolders and symlinks.
```
taoeffect, line 31: Please restore this comment and remove the eslint-disable thing

src/hash.ts
```diff
@@ -5,9 +5,9 @@ import { createEntryFromFile, multicodes } from './utils.ts'
 // TODO: use https://doc.deno.land/https://deno.land/std@0.140.0/streams/mod.ts/~/iterateReader instead to read in large files, and then use blake2b[Init,Update,Final] to iteratively calculate the hash
 //       Verify that it returns the same hash as when we use readAll https://doc.deno.land/https://deno.land/std@0.140.0/streams/mod.ts/~/readAll
 
-export async function hash (args: string[], multicode: number = multicodes.RAW, internal = false) {
+export async function hash (args: string[], multicode: number = multicodes.RAW, internal = false): Promise<string> {
   const [filename] = args
-  if (!filename) {
+  if (filename === undefined || filename === '') {
```
taoeffect, line 10: here

src/database-fs.ts
corrideat: `!!entry.isFile` would be simpler but I'm not sure why this is (was) even an issue here, considering that according to the [docs](https://docs.deno.com/api/deno/~/Deno.DirEntry) this should be a boolean to begin with.

scripts/build.ts
```diff
@@ -28,18 +28,19 @@ const options = {
...
+    'npm:@multiformats/blake2@1.0.13',
+    'npm:@chelonia/crypto@1.0.1'
   ],
   format: 'esm',
   outdir: 'build',
   splitting: false
 }
 const result = await esbuild.build(options)
-if (result.errors.length) {
+if ((result.errors as esbuild.Message[]).length > 0) {
```
corrideat, line 39: Not sure why this cast is needed, since the type should be implied (OTOH, I believe you might need a cast on `options` in line 37 to add `as esbuild.BuildOptions`)

src/database-fs.ts
```diff
@@ -20,26 +20,33 @@ export async function clear (): Promise<void> {
...
 }
 
-export async function readData (key: string): Promise<Uint8Array | string | void> {
-  // Necessary here to thwart path traversal attacks.
+export async function readData (key: string): Promise<Uint8Array | string | undefined> {
   checkKey(key)
-  return await Deno.readFile(path.join(dataFolder, key))
-    .catch(_err => undefined)
+  try {
+    // assign to a variable before returning to satisfy no-return-await
```
corrideat: Maybe @taoeffect will have a different take, but I'm not a big fan of comments such as these, unless you're doing something weird and/or they're used sparingly. The reason is that to keep the signal-to-noise ratio high, since referring to a specific lint rule doesn't help understand the code much.

In this case, I would find it more useful, if you wanted to document why you made this change, if you'd add a GitHub comment on your PR.

src/deploy.ts
corrideat: I agree with @taoeffect here, but like I said earlier, I don't think the issue is so much the rules but the configuration.

src/eventsAfter.ts
corrideat: Not only is this weird, but it's also wrong, as you can only call `.text()` once (because it consumes the response body).

corrideat (changes requested):
This is a good start, but it looks like there are several major issues, many of which are unnecessary weirdness added to the code for seemingly no reason.

Trying these changes locally, `npm install` doesn't seem to work due to dependency issues, and so I couldn't verify whether linting works locally.

I'd revert some/most/all of the changes that add weirdness, but before disabling the rule please try to see why the linter is complaining. In many cases, the changes don't seem needed even if the rule is strictly applied, because the type is satisfied (for example, `entry.isFile` should always be a boolean, or `path.join` should always be a string).

eslint.config.js
```diff
@@ -0,0 +1,56 @@
...
+      '@typescript-eslint/no-unnecessary-type-assertion': 'off',
+      '@typescript-eslint/no-unnecessary-condition': 'off', 
+      '@typescript-eslint/no-unnecessary-type-conversion': 'off',
+      '@typescript-eslint/require-await': 'off',
+      '@typescript-eslint/only-throw-error': 'off',
+      '@typescript-eslint/consistent-indexed-object-style': 'off',
+      '@typescript-eslint/use-unknown-in-catch-callback-variable': 'off',
+      'no-constant-binary-expression': 'off',
+      'arrow-body-style': 'off',
+      'no-console': 'off',
```
taoeffect, line 53: I do not see anything related to `standard` here. So where is `standard` enforced?

src/database-fs.ts
akhileshthite: I reverted this in the latest commits.

src/database-sqlite.ts
```diff
@@ -31,7 +31,7 @@ export async function initStorage (options: Record<string, unknown> = {}): Promi
   // Important: keep this in sync with the schema used in GroupIncome.
   db.run('CREATE TABLE IF NOT EXISTS Data(key TEXT NOT NULL PRIMARY KEY, value TEXT NOT NULL)')
   dbPath = filepath
-  if (!options.internal) {
+  if (options.internal !== true) {
```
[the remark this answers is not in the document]
akhileshthite: It's not misconfiguration, I've answered this on slack.

akhileshthite:
> This PR doesn't enforce `standard`. You can verify by adding a semi-colon anywhere and `deno task lint` will still pass. Please make sure that `standard` rules are being enforced.

Previously, we used `.eslintrc.json` with:
```json
{ "extends": "ts-standard" }
```

This relied on `ts-standard`, which wraps ESLint with Standard-style rules + TypeScript support. However:
- `.eslintrc.*` (the legacy ESLint config format) is now officially deprecated:
https://eslint.org/docs/latest/use/configure/configuration-files-deprecated
- `ts-standard` itself isn't actively maintained or updated to support modern ESLint Flat config: https://github.com/standard/ts-standard

Other options we evaluated:
- `eslint-config-standard` requires `.eslintrc.*`, which is also deprecated.
- Like groupincome, adding ESLint config to package.json is no longer supported:
https://eslint.org/docs/latest/use/configure/migration-guide#packagejson-configuration-no-longer-supported

Latest Recommendation:
- The recommended way to configure ESLint is to use Flat config via `eslint.config.js`:
https://eslint.org/docs/latest/use/configure/configuration-files
- `eslint-config-love` is the most popular config for a Standard-style + TypeScript setup that works with Flat config.
One limitation we hit is that `eslint-config-love` does not enforce the stylistic rules that `ts-standard` did — such as no semicolons, single quotes, spacing, etc. **So we still need to extend it with explicit rules to enforce formatting.**

I’ve included the `ts-standard` specific rules in our ESLint config and tested applying them using the `--fix` flag.
https://github.com/okTurtles/chel/pull/51/commits/c05ae54b216243c9fbc9959abefb9b700e065f93
Should we include the `--fix` flag by default in the `deno.json` lint task?

Please let me know what you’d like me to do!

eslint.config.js
akhileshthite: https://github.com/okTurtles/chel/pull/51#issuecomment-3092573653

src/database-sqlite.ts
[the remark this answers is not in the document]
corrideat: > In the attached image here, this is what we get if the strict-boolean-expressions rule is off (comparing current and previous setup). This is not any misconfiguration, it’s just TS being weird

It's not TS being weird, and in this specific instance it's not eslint misconfiguration, but it's 'code' misconfiguration. If you see, there's `initStorage (options: Record<string, unknown> = {}):`

That type is wrong, since `options` is too generic in this case. Ideally, we'd have a type here with all possible options, and `internal` would be there. Now, it may make sense for `internal` to be optional, and in this case `@typescript-eslint/strict-boolean-expressions` would still need `=== true`.

So, summing up, I agree mostly with that `@typescript-eslint/strict-boolean-expressions` should be disabled because it doesn't even accept `!!var` or `!var` as explicit boolean casting (and I don't see much of the point of mandating `=== true/false` checks or `?? true/false`). However, in this case, this lint issue caught a different, related, issue that should be addressed (the non-specificity of `options`).

corrideat:
> EDIT: It seems npm install isn't meant to be used with this, and the command is deno install.

Insofar as `chel` is an `npm` package, `npm install` (or, at least, `npm install --production`) should work lest it becomes uninstallable using `npm`. I tested this with `npm pack` and `npm install` and, however, it seems to work for some reason. Still, I see very little reason to have something like `eslint-config-love` or `eslint` as a `dependency` (as opposed to as a `devDependency`).

eslint.config.js
corrideat: I haven't yet looked at each of them, so I don't have a list. Maybe the most time-efficient approach here is keeping it as is and working on removing as many as possible from here on a separate issue.

taoeffect: ok, then @akhileshthite can you remove the two @corrideat mentioned if they don't uglify the code too much? `@typescript-eslint/only-throw-error` should not affect the code at all. Please check if there are any other rules that would not uglify the code at all either, and remove those too.

src/get.ts
```diff
@@ -11,7 +11,7 @@ chel get https://url.com mygreatlongkey > file.png
 import { flags, writeAll } from './deps.ts'
 import { exit, getBackend, isURL, readRemoteData } from './utils.ts'
 
-export async function get (args: string[]): Promise<void> {
```
taoeffect, line 14: All of these "space around functions" changes need to be reverted. They seem to violate standard too, check `group-income`, doing this results in:

```
225:20  error  Missing space before function parentheses  space-before-function-paren
```

src/upload.ts
```diff
@@ -21,17 +21,17 @@ export async function upload (args: string[], internal = false) {
...
-          // raw file type
-          break
-        case 'm':
-          type = multicodes.SHELTER_CONTRACT_MANIFEST
-          break
-        case 't':
-          type = multicodes.SHELTER_CONTRACT_TEXT
-          break
-        default:
-          throw new Error('Unknown file type: ' + filepath_[0])
```
taoeffect, line 34: These changes need to be reverted too, what rule is causing this?

src/verifySignature.ts
```diff
@@ -35,10 +69,10 @@ export const verifySignature = async (args: string[], internal = false) => {
     return keyId(k) === manifest.signature.keyId
   })
 
-  if (externalKeyDescriptor) {
+  if (externalKeyDescriptor !== null) {
```
taoeffect, line 72: here

src/utils.ts
```diff
@@ -71,9 +71,9 @@ export async function getBackend (src: string, { type, create } = { type: '', cr
 
   let initOptions
   switch (from) {
-    case 'fs': initOptions = fsOptions; break
-    case 'sqlite': initOptions = sqliteOptions; break
-    default: throw new Error(`unknown backend type: "${from}"`)
+  case 'fs': initOptions = fsOptions; break
+  case 'sqlite': initOptions = sqliteOptions; break
+  default: throw new Error(`unknown backend type: "${from}"`)
```
taoeffect, line 76: here

src/upload.ts
```diff
@@ -21,17 +21,17 @@ export async function upload (args: string[], internal = false) {
...
+        // raw file type
+        break
+      case 'm':
+        type = multicodes.SHELTER_CONTRACT_MANIFEST
+        break
+      case 't':
+        type = multicodes.SHELTER_CONTRACT_TEXT
+        break
+      default:
+        throw new Error('Unknown file type: ' + filepath_[0])
```
taoeffect, line 34: In fact, these changes should trigger a linter errors. Try this on `group-income`, it will result in a linter error. Can you make sure this results in a linter error here too?

src/upload.ts
```diff
@@ -47,40 +47,40 @@ export async function upload (args: string[], internal = false) {
...
+async function uploadEntryToURL([cid, buffer]: Entry, url: string): Promise<string> {
   const form = new FormData()
   form.append('hash', cid)
   form.append('data', new Blob([buffer]))
   return await fetch(`${url}/dev-file`, { method: 'POST', body: form })
     .then(handleFetchResult('text'))
     .then(r => {
       if (r !== `/file/${cid}`) {
-        throw new Error(`server returned bad URL: ${r}`)
+        throw new Error(`server returned bad URL: ${String(r)}`)
```
taoeffect, line 58: Can this be fixed not with `String(r)` coercion, but by typing `r` on line 56? E.g. `.then(r: string`?

src/eventsAfter.ts
```diff
@@ -50,8 +50,8 @@ async function getMessagesSince (src: string, contractID: string, sinceHeight: n
   let currentHeight: number
   while (true) {
     const entry = await getMessage(currentHEAD)
-    if (!entry) {
-      throw new Deno.errors.NotFound(`entry ${currentHEAD} no longer exists.`)
+    if (entry === undefined) {
```
taoeffect, line 53: Missed this, please revert

src/database-fs.ts
```diff
@@ -35,11 +37,10 @@ export async function * iterKeys (): AsyncGenerator<string> {
   }
 }
 
+// deno-lint-ignore require-await
 export async function readData (key: string): Promise<Uint8Array | string | void> {
-  // Necessary here to thwart path traversal attacks.
   checkKey(key)
-  return await Deno.readFile(path.join(dataFolder, key))
-    .catch(_err => undefined)
+  return Deno.readFile(path.join(dataFolder, key)).catch(() => undefined)
```
taoeffect, line 43: This would not need an `// deno-lint-ignore require-await` comment if the function were not marked `async`.

So the `async` label can simply be removed, and then you should be able to remove the comment too

eslint.config.js
```diff
@@ -0,0 +1,72 @@
+import parser from '@typescript-eslint/parser'
+import tseslint from '@typescript-eslint/eslint-plugin'
+import love from 'eslint-config-love'
+import eslintComments from 'eslint-plugin-eslint-comments'
+
+export default [
+  {
+    ...love,
```
taoeffect, line 8: While trying to get this to work with Zed, and looking at this file in this PR for Zed: https://github.com/VortechSolutions/gridjs-svelte/pull/16/files#diff-a32a0887ed9d1d707bbb3b845b7df7fd40e673c47e7b60a3ebd896b68d3b8839

I'm noticing a few things here:

1. It uses `ts.config(` instead of `export default`
2. If you look up `'@typescript-eslint/eslint-plugin'` on npm, the github actually redirects you to `typescript-eslint` (which is what that Zed PR uses)
3. [Their documentation recommends](https://typescript-eslint.io/getting-started#step-2-configuration) the use of `export default tseslint.config(`, which we're not using here
4. The example in the Zed PR doesn't use `import parser from '@typescript-eslint/parser'` and neither does the documentation, so do we need it here?

Their documentation [goes into more details about what [`tseslint.config` actually means](https://typescript-eslint.io/packages/typescript-eslint#config). Notice that it uses also:

```
  eslint.configs.recommended,
  tseslint.configs.recommended,
```

Perhaps we should use that too instead of this `'eslint-config-love'` plugin?

I'm just looking to see if there are any ways to simplify this config a bit, as it seems quite long and is doing stuff that the neither the Zed example PR nor the docs are doing.

eslint.config.js
```diff
@@ -0,0 +1,72 @@
+import parser from '@typescript-eslint/parser'
+import tseslint from '@typescript-eslint/eslint-plugin'
+import love from 'eslint-config-love'
+import eslintComments from 'eslint-plugin-eslint-comments'
```
taoeffect, line 4: Do we need this plugin?

src/version.ts
```diff
@@ -1,4 +1,4 @@
-export function version () {
+export function version (): void {
   // @ts-expect-error: __build__ is injected at build time
```
corrideat, line 2: Instead of `@ts-expect-error`, can we please define `__build__` in a definition file?

src/verifySignature.ts
```diff
@@ -38,7 +72,7 @@ export const verifySignature = async (args: string[], internal = false) => {
   if (externalKeyDescriptor) {
     const id = keyId(externalKeyDescriptor.pubkey)
     if (manifest.signature.keyId !== id) {
-      exit(`Invalid manifest signature: key ID doesn\'t match the provided key file. Expected ${id} but got ${manifest.signature.keyId}.`, internal)
+      exit(`Invalid manifest signature: key ID doesn't match the provided key file. Expected ${String(id)} but got ${String(manifest.signature.keyId)}.`, internal)
```
corrideat, line 75: Can we please revert these changes? Key IDs are guaranteed to be strings.

src/upload.ts
```diff
@@ -5,10 +5,10 @@ import { type Entry, createEntryFromFile, isDir, multicodes, revokeNet } from '.
 
 // chel upload <url-or-dir-or-sqlitedb> <file1> [<file2> [<file3> ...]]
 
-export async function upload (args: string[], internal = false) {
+export async function upload (args: string[], internal = false): Promise<Array<[string, string]>> {
```
corrideat, line 8: Pesonally, I prefer `[string, string][]` over `Array<[string, string]>`.

src/manifest.ts
```diff
@@ -5,26 +5,30 @@
...
+  const parsedFilepath = path.parse(contractFile)
   const { name: contractFileName, base: contractBasename, dir: contractDir } = parsedFilepath
   const name = parsedArgs.name || parsedArgs.n || contractFileName
   const version = parsedArgs.version || parsedArgs.v || 'x'
   const slim: string | undefined = (parsedArgs.slim || parsedArgs.s) as string | undefined
   const outFile: string = (parsedArgs.out as string) || path.join(contractDir, `${contractFileName}.${version}.manifest.json`)
   if (!keyFile) exit('Missing signing key file')
 
-  const signingKeyDescriptor = await readJsonFile(keyFile)
+  const signingKeyDescriptor = await readJsonFile(keyFile) as { privkey: string }
```
corrideat, line 31: Might be a good idea to use a runtime check here instead of a type cast.

src/deploy.ts
```diff
@@ -9,13 +9,18 @@ import { upload } from './upload.ts'
...
+
+export async function deploy (args: string[]): Promise<void> {
   const [urlOrDirOrSqliteFile, ...manifests] = args
   if (manifests.length === 0) throw new Error('missing url or manifests!')
   const toUpload = []
   for (const manifestPath of manifests) {
-    const json = JSON.parse(Deno.readTextFileSync(manifestPath))
-    const body = JSON.parse(json.body)
+    const json = JSON.parse(Deno.readTextFileSync(manifestPath)) as { body: string }
+    const body = JSON.parse(json.body) as ContractBody
```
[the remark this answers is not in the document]
taoeffect: FYI, if you want runtime checks for these, consider using [zod](https://zod.dev/), because it will, AFAIU, do both static typing + runtime checking

src/version.ts
akhileshthite: Tried all options to get TypeScript to recognize `__build__` – including updating `tsconfig.json` with correct `include` and `typeRoots` paths – but still getting `TS2304: Cannot find name '__build__'` on `deno task test`.

Switching to `import './types/build.d.ts'` as a reliable workaround instead of using `// @ts-expect-error`.

taoeffect (approved):
Fantastic work @akhileshthite!

Approved!

package.json
```diff
@@ -23,7 +23,7 @@
     "axios": "0.27.2",
     "rimraf": "3.0.2",
     "tar": "6.1.11",
-    "zod": "^4.0.5"
```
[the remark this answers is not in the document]
akhileshthite: I installed it with `deno install npm:zod`.

package.json
```diff
@@ -22,11 +22,11 @@
   "dependencies": {
     "axios": "0.27.2",
     "rimraf": "3.0.2",
-    "tar": "6.1.11"
+    "tar": "6.1.11",
+    "zod": "4.0.5"
```
corrideat, line 26: I'm not really sure why `zod` is here. I don't think it should be, since it's a runtime dependency. Instead, it should be imported like all of the other dependencies using `deps.ts`.

PR #58, opened 2025-07-23
replace npx with deno

taoeffect:
Do not use `npx` — ever.

This helps protect against supply-chain attacks.

Screenshot 2025-07-23 at 10 42 45 AM

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "tasks": {
     "chel": "deno task lint && deno task build && deno run --allow-net --allow-read=. --allow-write=. build/main.js",
-    "lint": "npx eslint ./src ./scripts",
+    "lint": "deno run --allow-env --allow-read --allow-sys --deny-sys=\"networkInterfaces\" npm:eslint@9.31.0 ./src ./scripts",
     "compile": "deno run --allow-run --allow-read=. --allow-write=./dist scripts/compile.ts",
     "build": "deno run --node-modules-dir --allow-run --allow-read --allow-env --allow-write=./build --allow-net scripts/build.ts",
     "dist": "deno lint . && deno task lint && deno task build && deno task compile",
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "tasks": {
     "chel": "deno task lint && deno task build && deno run --allow-net --allow-read=. --allow-write=. build/main.js",
-    "lint": "deno run --allow-env --allow-read --allow-sys --deny-sys=\"networkInterfaces\" npm:eslint@9.31.0 ./src ./scripts",
+    "lint": "deno run --allow-env --allow-read=. --allow-sys --deny-sys=\"networkInterfaces\" npm:eslint@9.31.0 ./src ./scripts",
     "compile": "deno run --allow-run --allow-read=. --allow-write=./dist scripts/compile.ts",
     "build": "deno run --node-modules-dir --allow-run --allow-read --allow-env --allow-write=./build --allow-net scripts/build.ts",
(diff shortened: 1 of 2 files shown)


deno.json
```diff
@@ -1,11 +1,11 @@
...
   "tasks": {
     "chel": "deno task lint && deno task build && deno run --allow-net --allow-read=. --allow-write=. build/main.js",
-    "lint": "npx eslint ./src ./scripts",
+    "lint": "deno run --allow-env --allow-read=. --allow-sys --deny-sys=\"networkInterfaces\" npm:eslint@9.31.0 ./src ./scripts",
     "compile": "deno run --allow-run --allow-read=. --allow-write=./dist scripts/compile.ts",
-    "build": "deno run --node-modules-dir --allow-run --allow-read --allow-env --allow-write=./build --allow-net scripts/build.ts",
+    "build": "deno run --node-modules-dir --allow-run --allow-read=. --allow-env --allow-write=./build --allow-net scripts/build.ts",
     "dist": "deno lint . && deno task lint && deno task build && deno task compile",
-    "test": "deno task lint && deno test --allow-read --allow-write --allow-env --allow-ffi --allow-net"
+    "test": "deno task lint && deno test --allow-read=.,$HOME/.cache --allow-write=.,$HOME/.cache --allow-env --allow-ffi --allow-net"
```
akhileshthite, line 8: Getting `throw new Error("Failed to load SQLite3 Dynamic Library", { cause: e });` on `deno task test`. It says `Caused by: NotCapable: Requires read access to "/deno/plug", run again with the --allow-read flag`

akhileshthite (approved):

corrideat:
>  will sandbox the NPM packages, so if they're malicious, their damage is mitigated unlike when run via npx.

Fair enough. AFAICT there's no out-of-the-box sandbox for with `npx` (Node has it, but configuring it is more complicated over just using Deno, which we're already using).

However, that wasn't entirely the reason given on the PR (or at least, the screenshots implied to me some kind of protection against attacks done through npmjs.com credential stealing).

That said, if you're concerned about this kind of attack, I'd suggest opening an issue for refining the sandbox permissions.

For example, `--allow-env --allow-read --allow-sys --deny-sys=\"networkInterfaces\"` is too liberal. I think it _may_ have helped against some kind of RAT (as per the screenshots) by neutralising it or making it less powerful, but those permissions don't seem nearly good enough for other kind of malicious behaviour, such as credential stealing, cryptominers (the mining itself likely can't be stopped, but blocking network communication might break it or at least make it unprofitable), etc.

corrideat:
> These permissions are more than enough to stop the attack mentioned in the thread AFAICT. 

Yeah, I misread `--allow-env` for `--allow-net` somehow.

> [...] stop the attack mentioned [...]

Ideally we'd want to stop all known classes of attacks that can be stopped, but disallowing both write and network permissions seems to be good enough. I'd further restrict `--allow-read`, but this permission isn't too useful alone, without being able to write or to communicate.

taoeffect:
> I'd further restrict --allow-read, but this permission isn't too useful alone, without being able to write or to communicate.

I tried to but it wouldn't let `eslint` run on macOS if I did `--allow-read=.` as eslint kept traversing upwards looking for `eslint.config.js` files, and blocking it prevented it from running at all. :-\

corrideat:
> eslint kept traversing upwards looking for eslint.config.js files, and blocking it prevented it from running at all. :-\

I wonder is this is configurable. Either way, I don't see it as a high priority issue as I can't see (maybe I'm unimaginative) what use reading things without being able to write or send them over to a network is.

Issue #37, opened 2024-03-28
Prevent `npm install` from deleting `bin/chel`


taoeffect:
### Problem

Running `npm install` for some reason deletes the `bin/chel` file and replaces it with the `chel` binary.

This prevents the package from installing the binary when it's installed later by users.

### Solution

Figure out why it does this and prevent it.

corrideat:
Relevant question: <https://stackoverflow.com/questions/54212147/npm-run-postinstall-script-only-when-installed-as-dependency>

This has to do with how `npm` lifecycle scripts work, and because `npm` doesn't have separate ways of handling installing on the package itself (`npm install`) and as a dependency (`npm install <package-name>`).

The solution to this seems to detect what it is (e.g., by having a file that's part of the repo but not of the package when install as a dependency) and run it.

PR #62, opened 2025-07-29
refactor: replace axios with native fetch for binary download

harshjainh2j:
refactor: replace axios with undici for binary download

Removed axios as a dependency and replaced it with undici's fetch API to download and extract the binary file. This reduces package bloat and improves compatibility with Node.js versions below 18.

- Uses undici's fetch for HTTP requests
- Streams the response directly into tar extractor
- Preserves existing error handling and logging
- Tested to ensure proper binary installation


scripts/binary-install.js
```diff
@@ -102,4 +108,4 @@ class Binary {
   }
 }
 
-module.exports.Binary = Binary
+module.exports.Binary = Binary
```
taoeffect, line 111: Ensure newline at EOF

scripts/binary-install.js
```diff
@@ -7,8 +7,9 @@ const process = require('node:process')
 const { existsSync, mkdirSync } = require('fs')
 const { join, resolve } = require('path')
 const { spawnSync } = require('child_process')
+const { Readable } = require('node:stream') // Required for stream conversion
+const { fetch } = require('undici') // Keep undici for fetch
```
taoeffect, line 11: Use built-in `fetch`

bin/chel
corrideat, line 1: This shouldn't be deleted

package.json
```diff
@@ -20,9 +20,9 @@
     "scripts/**/*"
   ],
   "dependencies": {
-    "axios": "0.27.2",
     "rimraf": "3.0.2",
     "tar": "6.1.11",
+    "undici": "7.12.0",
```
taoeffect, line 25: Remove this dependency

bin/chel.exe
corrideat, line 1: I guess it's fine to add this file, but it should just be a placeholder.

scripts/binary-install.js
corrideat, line 11: `fetch` should be defined without using `undici`, so this line and the new `undici` can be removed.

taoeffect (changes requested):
Thanks for the PR @harshjainh2j! I left some feedback.

Also, sorry about this, but please avoid running `npm install`, as there's an open issue #37 that causes a local run of `npm install` to delete `bin/chel` and replace it with a binary - so those changes here need to be reverted.

scripts/binary-install.js
corrideat, line 111: Missing end-of-line

package.json
corrideat, line 25: As mentioned, this can be removed.

corrideat (changes requested):
Thanks a lot for your help @harshjainh2j! The PR looks fine, but please note my minor comments.

corrideat:
> Also, sorry about this, but please avoid running npm install, as there's an open issue https://github.com/okTurtles/chel/issues/37 that causes a local run of npm install to delete bin/chel and replace it with a binary - so those changes here need to be reverted.'

I'm afraid `npm install` is needed in this case since `package-lock.json` is modified, but maybe try running `npm install --ignore-scripts`, or manually verify the contents of `bin`.

scripts/binary-install.js
```diff
@@ -64,23 +65,28 @@ class Binary {
...
-        error(`Error fetching release: ${e.message}`)
+    try {
+      const response = await fetch(this.url, fetchOptions)
+
+      if (!response.ok) {
+        throw new Error(`Failed to download: ${response.status} ${response.statusText}`)
+      }
+
+      // Convert web stream to Node.js stream (CRITICAL FIX)
+      const nodeStream = Readable.fromWeb(response.body)
```
corrideat, line 76: This line in itself is fine, but I've seen a few Node.js bugs related to using this specific API (I mean that the Node.js logic was, at some point at least, buggy, for converting from web streams to native streams).

harshjainh2j:
Hey ! @corrideat , @taoeffect , I have removed undici and used the global fetch instead . and ensured no accidental changes to bin/chel after npm install, and manually verified its contents.

taoeffect:
Thanks @harshjainh2j! I think you'll need to manually use some `git` commands to prevent these changes from showing up though, as `chel` is still being modified, and `chel.exe` is being added (keep in mind this will happen each time you run `npm install` until #37 is closed, so don't run that command):

Screenshot 2025-07-30 at 9 55 29 AM

harshjainh2j:
Hey!  @corrideat @taoeffect ,Thanks for the feedback  , I’ve made all the requested changes:

1. Added a line break at the end of scripts/binary-install.js

2. Restored bin/chel from master

3. Replaced bin/chel.exe with the same placeholder as bin/chel

corrideat (approved):

taoeffect (approved):
Alright! Nicely done @harshjainh2j! 

Testing this will require trying out an update I suppose, so it's unclear whether this will work, but it looks like it should so I'll approve and merge.

Thanks for the PR! 😄

corrideat:
> Testing this will require trying out an update

Not really. Running `npm install` should be a sufficient test, or, alternatively, `npm pack`, and then `npm install <path-to-package>` on a different project.

PR #66, opened 2025-08-15
feat: add TodoMVC example with identity contracts and KV store

akhileshthite:
https://github.com/user-attachments/assets/a2a445af-2a79-4145-9c0f-d8a49a852295

### DIFF
--- a/todomvc/assets/js/app.js
+++ b/todomvc/assets/js/app.js
@@ -1,40 +1,79 @@
-// TodoMVC App using Chelonia (no service worker)
-// Based on pre-SW Group Income architecture
-
-const sbp = (function() {
-  const selectors = {}
-  
-  function sbp(selector, ...args) {
-    if (selectors[selector]) {
-      return selectors[selector](...args)
-    }
-    throw new Error(`Unknown selector: ${selector}`)
-  }
-  
-  sbp.selectors = {
-    register: (selectorMap) => {
-      Object.assign(selectors, selectorMap)
-    }
-  }
-  
-  return sbp
-})()
-
-// Simplified storage for demo (in real app this would be Chelonia contracts)
-const demoStorage = {
-  contracts: {},
-  kvStore: {}
-}
+// TodoMVC App using Chelonia with CDN imports
 
 class TodoMVCApp {
   constructor() {
-    this.contractID = null
-    this.currentFilter = 'all'
     this.todos = {}
+    this.filter = 'all'
+    this.contractID = null
     
-    this.initializeElements()
-    this.bindEvents()
-    this.setupRouting()
+    // Wait for SBP/Chelonia to be available
+    this.waitForLibraries().then(() => {
+      this.initializeElements()
+      this.bindEvents()
+      this.render()
+      
+      // Check for existing login
+      this.checkExistingLogin()
+    })
+  }
+
+  async waitForLibraries() {
+    // Wait for SBP to be loaded from CDN
+    let attempts = 0
+    const maxAttempts = 50 // 5 seconds max wait
+    
+    while (!window.sbp && attempts < maxAttempts) {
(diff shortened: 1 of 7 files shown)


todomvc/assets/js/app.js
```diff
@@ -0,0 +1,440 @@
+// TodoMVC App using Chelonia (no service worker)
+// Based on pre-SW Group Income architecture
+
+const sbp = (function() {
```
taoeffect, line 4: Not sure why you're defining sbp. SBP must be imported

todomvc/assets/js/app.js
```diff
@@ -0,0 +1,440 @@
...
+    // Show/hide main and footer sections
+    const hasTodos = todos.length > 0
+    this.mainSection.style.display = hasTodos ? 'block' : 'none'
+    this.footerSection.style.display = hasTodos ? 'block' : 'none'
+    
+    // Update toggle all checkbox
+    this.toggleAllCheckbox.checked = activeTodos.length === 0 && todos.length > 0
+    
+    // Render todo list
+    this.todoList.innerHTML = filteredTodos.map(todo => this.renderTodoItem(todo)).join('')
```
corrideat, line 307: I realise this is a demo app, but still `.innerHTML` is a code smell. There are at least two ways you can address this. The 'classical' way, i.e., building DOM elements using `document.createElement` and the 'modern' way using the `<template>` element.

todomvc/assets/js/app.js
```diff
@@ -0,0 +1,440 @@
...
+      button.addEventListener('click', (e) => {
+        const li = e.target.closest('li')
+        const id = li.dataset.id
+        this.deleteTodo(id)
+      })
+    })
+    
+    // Edit todo
+    this.todoList.querySelectorAll('label').forEach(label => {
+      label.addEventListener('dblclick', (e) => {
```
corrideat, line 355: I realise this comes from the TODO MVC examples, but using `dblclick` exclusively makes it mostly unusable in mobile and for keyboard users.

corrideat:
This looks like a good start. I left some accessibility-related comments that, while they are issues in the code this is based on (i.e., the other TODO MVC examples), there's maybe little reason to replicate here.

I think that to simplify things, up to you, it's fine to use a CDN for importing `sbp` and `@chelonia/lib` here, since this isn't supposed to be a security-sensitive app.

I see that you're adding login functionality. This part will likely be the most complex aspect of this MVC, since login currently takes several steps, but it'll also be a good opportunity to see how simple we can make it.

akhileshthite:
he TodoMVC uses SBP (@sbp/sbp@2.4.1) and Chelonia (@chelonia/lib@1.2.2) libraries loaded from CDN, with custom `todomvc/*` selectors that provide browser-compatible identity and KV operations while following SBP selector patterns and Chelonia architecture.

corrideat:
I've taken a look at this again, and these are some thoughts:

1. It's fine if you don't use any frameworks (and I'd say the same if you were using a framework), but the current `app.js` file is already too complicated. The point of this PR is to serve as a starting point / example / demo for Chelonia, so ideally we'd want the Chelonia-relevant parts to be self-contained and defined separately from the UI. In other words, the audience would be application developers who probably already know how to implement a web app, so they could gloss over the non-Chelonia parts and focus on the Chelonia parts.
2. I don't see this PR using any contracts whatsoever (right now, anyhow), which it should, since that's a fundamental aspect of Chelonia. That also means that state will need to be loaded from a contract.

`chelonia.db/set` and `chelonia.db/get` should in fact work on simple strings, but colons have a special meaning. However, you probably shouldn't be using these selectors (directly) on this app.

taoeffect:
Chelonia has changed a lot, so closing this stale PR

Issue #79, opened 2025-10-27
Implement `chel migrate`


taoeffect:
### Problem

A followup issue to #78 - we need the ability for sysadmins to migrate their databases from one database to another.

### Solution

Implement a chel migrate command that takes two database arguments and migrates all data from one to the other.

Make sure to handle various edge cases like if a previous migration was done but cut short with `^C` and needs to be redone.

EDIT: please also remove the `database-*` files under `src` ([see comment](https://github.com/okTurtles/chel/pull/82#issuecomment-3458393355)).

corrideat:
> To be clear, this migrate command should work as copy, not move. It shouldn't be a destructive command.

Not disagreeing with you, just pointing out that migration can result in a messy state in certain conditions.

> Yeah, I suppose let's do that.

Should we use `chelonia.db` for one of the backends and direct import for the other, or direct import for both? If one and one, which should use `chelonia.db`? I'd say use `chelonia.db` for the current (origin) backend and direct import for the new (target) backend.

PR #82, opened 2025-10-28
Feat/redis

### LINKED ISSUE #78: Redis (or compatible) db backend
_(this issue is paired with: https://github.com/okTurtles/group-income/issues/2920)_

### Problem

The sqlite database is occasionally resulting in forked chains when the server is upgraded because it seems like the sqlite database is not getting saved properly, or something. So an older version of the database is being restored on the new version, requiring people to re-login.

Redis is a well-known persistent KV database, although there was some drama and there are now various community forks of it:

- https://github.com/dragonflydb/dragonfly
- https://docs.keydb.dev/
- https://valkey.io/

### Solution

Pick one of the Redis forks and implement it for our backend.

[Dragonfly seems to be the fastest](https://www.dragonflydb.io/blog/dragonfly-vs-valkey-benchmark-on-google-cloud) and is supposedly compatible with Redis libraries.

The `'redis'` or `'ioredis'` libraries should be compatible with it, give them a try.

We should also add support to the `chel` command to add support for migrating between databases.

corrideat:
Closes #78

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -58,18 +58,18 @@ function assertArgs(path, suffix) {
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/_common/from_file_url.ts
-function assertArg(url2) {
-  url2 = url2 instanceof URL ? url2 : new URL(url2);
-  if (url2.protocol !== "file:") {
-    throw new TypeError(`URL must be a file URL: received "${url2.protocol}"`);
+function assertArg(url) {
+  url = url instanceof URL ? url : new URL(url);
+  if (url.protocol !== "file:") {
+    throw new TypeError(`URL must be a file URL: received "${url.protocol}"`);
   }
-  return url2;
+  return url;
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/posix/from_file_url.ts
-function fromFileUrl(url2) {
-  url2 = assertArg(url2);
-  return decodeURIComponent(url2.pathname.replace(/%(?![0-9A-Fa-f]{2})/g, "%25"));
+function fromFileUrl(url) {
+  url = assertArg(url);
+  return decodeURIComponent(url.pathname.replace(/%(?![0-9A-Fa-f]{2})/g, "%25"));
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/_common/strip_trailing_separators.ts
@@ -126,11 +126,11 @@ function isWindowsDeviceRoot(code2) {
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/windows/from_file_url.ts
-function fromFileUrl2(url2) {
-  url2 = assertArg(url2);
-  let path = decodeURIComponent(url2.pathname.replace(/\//g, "\\").replace(/%(?![0-9A-Fa-f]{2})/g, "%25")).replace(/^\\*([A-Za-z]:)(\\|$)/, "$1\\");
-  if (url2.hostname !== "") {
-    path = `\\\\${url2.hostname}${path}`;
+function fromFileUrl2(url) {
+  url = assertArg(url);
+  let path = decodeURIComponent(url.pathname.replace(/\//g, "\\").replace(/%(?![0-9A-Fa-f]{2})/g, "%25")).replace(/^\\*([A-Za-z]:)(\\|$)/, "$1\\");
+  if (url.hostname !== "") {
+    path = `\\\\${url.hostname}${path}`;
   }
   return path;
 }
@@ -349,8 +349,8 @@ function extname3(path) {
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/from_file_url.ts
-function fromFileUrl3(url2) {
-  return isWindows ? fromFileUrl2(url2) : fromFileUrl(url2);
+function fromFileUrl3(url) {
+  return isWindows ? fromFileUrl2(url) : fromFileUrl(url);
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/posix/is_absolute.ts
@@ -1080,9 +1080,9 @@ function toFileUrl(path) {
   if (!isAbsolute(path)) {
     throw new TypeError(`Path must be absolute: received "${path}"`);
   }
-  const url2 = new URL("file:///");
-  url2.pathname = encodeWhitespace(path.replace(/%/g, "%25").replace(/\\/g, "%5C"));
-  return url2;
+  const url = new URL("file:///");
+  url.pathname = encodeWhitespace(path.replace(/%/g, "%25").replace(/\\/g, "%5C"));
+  return url;
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/windows/to_file_url.ts
@@ -1091,15 +1091,15 @@ function toFileUrl2(path) {
     throw new TypeError(`Path must be absolute: received "${path}"`);
   }
   const [, hostname2, pathname] = path.match(/^(?:[/\\]{2}([^/\\]+)(?=[/\\](?:[^/\\]|$)))?(.*)/);
-  const url2 = new URL("file:///");
-  url2.pathname = encodeWhitespace(pathname.replace(/%/g, "%25"));
+  const url = new URL("file:///");
+  url.pathname = encodeWhitespace(pathname.replace(/%/g, "%25"));
   if (hostname2 !== void 0 && hostname2 !== "localhost") {
-    url2.hostname = hostname2;
-    if (!url2.hostname) {
-      throw new TypeError(`Invalid hostname: "${url2.hostname}"`);
+    url.hostname = hostname2;
+    if (!url.hostname) {
+      throw new TypeError(`Invalid hostname: "${url.hostname}"`);
     }
   }
-  return url2;
+  return url;
 }
 
 // deno:https://jsr.io/@std/path/1.1.1/to_file_url.ts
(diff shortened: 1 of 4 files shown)


taoeffect:
It would be nice if all of the backends had the same type signatures… How else can they be easily switched?

corrideat:
Using `Buffer` AFAICT doesn't break anything, except maybe (but then not really, based on how we're using it) if using the `any:` key prefix.

corrideat:
> Is there a reason we aren't restricting all of the values to a specific type?

Can you elaborate? We support `string` or `Buffer`, and there's a good reason to allow `Buffer` (being able to store binary data)

corrideat:
> database-sqlite.ts

I believe you're looking at the wrong `database-sqlite.ts` file. The one under `src/serve` uses `Buffer`.

> Just because we've been doing something dumb doesn't mean we should keep doing it

Agreed that standardising on `Buffer` / `Uint8Array` would be better. However, this isn't so important since everything is then casted to a string, except if using `any:`. The reason for having multiple return types is (was), I imagine, to support various types (e.g., number, string, buffer, you-name-it), but, when catering to the lowest common denominator of backends (e.g., the `fs` backend), this isn't possible to do without in-band type annotations.

I can update the DB backends to return only `Buffer` / `Uint8Array`. Or should I update backends to contain type annotations (e.g., we can prefix string data with `'s'` and buffer data with `'b'`)?

corrideat:
> Oh boy, looks like we all missed the fact that there are database-* files all under src/. These should be removed if they're outdated. Can you handle that in this PR?

I didn't miss that, but was trying to keep the PR on focus (being porting GI backend), since the `database-*` files under `src` are slightly incompatible from the ones under `src/serve`. I agree that it should be dealt with, but probably not under this PR. I'd do it under #79 or maybe under a different 'DB backend cleanup' PR (#79 is relevant because the files under `src` implement some APIs that we need for migrate and also concerns `migrate` more generic).

> > but currently all strings seem to be returned as buffers, which is a bit different from how the other backends work.

> Is that really a problem?

Practically, no. I believe it's safe to always return a buffer. However, it _would be_ different from the other backends.

> I'm a bit hesitant to say that we should use string prefixes like s and b, as that wouldn't be a simple dropin upgrade to the currently running site.

Yeah, but it would sort of be if everything is prefixed as `b`. The reason I don't like annotations / prefixes is that for writing / reading it either requires two operations (one for reading / writing the type and another for data) or that it requires only one, but twice as much memory use for writing (we need to copy `input buffer` into `write buffer`, which is one octet longer).

corrideat:
> OK, then let's leave it as it is for now

Then this PR should be ready to review and merge.

taoeffect (commented):
This is looking really solid @corrideat! I was finally able to get this working locally with some effort and stumbling points.

I noticed however this error in the console:

```
{"level":50,"time":1761679459541,"pid":718743,"hostname":"fedora-linux-42","err":{"type":"TypeError","message":"Cannot read properties of undefined (reading 'byteLength')","stack":"TypeError: Cannot read properties of undefined (reading 'byteLength')\n    at U2 (file:///@chelonia/chel/build/main.js:81548:14)\n    at encryptPayload (file:///@chelonia/chel/build/main.js:81770:14)\n    at eventLoopTick (ext:core/01_core.js:179:7)\n    at async postEvent (file:///@chelonia/chel/build/main.js:81783:28)\n    at async _WebSocket.store-subscription (file:///@chelonia/chel/build/main.js:81846:13)\n    at async _WebSocket.push_action (file:///@chelonia/chel/build/main.js:100144:15)"},"msg":"[undefined] Failed to store subscription '206b9730-e6a7-88ef-9151-7cea3024c326' (updates.push.services.mozilla.com), removing it!"}
{"level":40,"time":1761679459542,"pid":718743,"hostname":"fedora-linux-42","err":{"type":"TypeError","message":"Cannot read properties of undefined (reading 'byteLength')","stack":"TypeError: Cannot read properties of undefined (reading 'byteLength')\n    at U2 (file:///@chelonia/chel/build/main.js:81548:14)\n    at encryptPayload (file:///@chelonia/chel/build/main.js:81770:14)\n    at eventLoopTick (ext:core/01_core.js:179:7)\n    at async postEvent (file:///@chelonia/chel/build/main.js:81783:28)\n    at async _WebSocket.store-subscription (file:///@chelonia/chel/build/main.js:81846:13)\n    at async _WebSocket.push_action (file:///@chelonia/chel/build/main.js:100144:15)"},"msg":"[undefined] Action 'store-subscription' for 'push_action' handler failed: Cannot read properties of undefined (reading 'byteLength')"}
{"level":50,"time":1761679481315,"pid":718743,"hostname":"fedora-linux-42","err":{"type":"TypeError","message":"Cannot read properties of undefined (reading 'byteLength')","stack":"TypeError: Cannot read properties of undefined (reading 'byteLength')\n    at U2 (file:///@chelonia/chel/build/main.js:81548:14)\n    at encryptPayload (file:///@chelonia/chel/build/main.js:81770:14)\n    at eventLoopTick (ext:core/01_core.js:179:7)\n    at async postEvent (file:///@chelonia/chel/build/main.js:81783:28)\n    at async _WebSocket.store-subscription (file:///@chelonia/chel/build/main.js:81846:13)\n    at async _WebSocket.push_action (file:///@chelonia/chel/build/main.js:100144:15)"},"msg":"[undefined] Failed to store subscription '71655f9f-7a26-867a-b051-5281a08e4f7d' (updates.push.services.mozilla.com), removing it!"}
```

It seems like it might be coming from on your libraries?

See the last line here:

```js
var init_encrypt2 = __esm({
  "node_modules/.deno/@apeleghq+rfc8188@1.0.7/node_modules/@apeleghq/rfc8188/dist/encrypt.mjs"() {
    R3 = async (e2, b, f, i2) => {
      let A2 = await globalThis.crypto.subtle.importKey("raw", b, "HKDF", false, ["deriveKey", "deriveBits"]), y = await globalThis.crypto.subtle.deriveKey({ name: "HKDF", hash: "SHA-256", info: e2.cek_info, salt: f }, A2, e2.params, false, i2), u2 = await globalThis.crypto.subtle.deriveBits({ name: "HKDF", hash: "SHA-256", info: e2.nonce_info, salt: f }, A2, e2.nonce_length << 3);
      return [y, function* () {
        let L = new ArrayBuffer(e2.nonce_length), c = new DataView(L), h2 = new Uint8Array(L), a = new Uint8Array(u2), g2 = 4294967295, o2 = (e2.nonce_length >> 2) - 1, s = new Array(o2).fill(0);
        for (; ; ) {
          for (let t = 0; t <= g2; t++) {
            c.setUint32(c.byteLength - 4, t, false);
            let n = new Uint8Array(e2.nonce_length);
            for (let r = 0; r < n.length; r++) n[r] = a[r] ^ h2[r];
            yield n;
          }
          for (let t = 0; t < o2; t++) {
            if (t === o2 - 1 && s[t] === g2) throw new RangeError("Maximum number of segments exceeded");
            if (s[t] = (s[t] + 1) % (g2 + 1), c.setUint32(c.byteLength - 4 * (t + 2), s[t], false), s[t] !== 0) break;
          }
        }
      }()];
    };
    E2 = R3;
    B3 = (e2) => ArrayBuffer.isView(e2) ? new Uint8Array(e2.buffer).subarray(e2.byteOffset, e2.byteOffset + e2.byteLength) : new Uint8Array(e2);
    w3 = B3;
    N10 = () => {
      let e2 = new Uint8Array(16);
      return globalThis.crypto.getRandomValues(e2), e2;
    };
    U2 = async (e2, b, f, i2, A2, y) => {
      if (f <= e2.tag_length + 1 || f > 4294967295) throw new RangeError("Invalid record size: " + f);
      if (i2.byteLength > 255) throw new RangeError("Key ID too long");
```

Which gets called by (last line here):

```js
    encryptPayload = async (subscription, data) => {
      const readableStream = new Response(data).body;
      if (!readableStream) throw new Error("Failed to create readable stream");
      const [asPublic, IKM] = await subscription.encryptionKeys;
      return K2(x2, readableStream, 32768, asPublic.buffer, IKM.buffer).then(async (bodyStream) => {
```

That in turn:

```js
    postEvent = async (subscription, event) => {
      const authorization = await vapidAuthorization(subscription.endpoint);
      const body = event ? await encryptPayload(subscription, event) : void 0;
```

taoeffect:
And one more issue found: I tried changing the port and not using `--network=host`:

```
sudo docker run -v $(pwd)/data/redis:/data -p 6380:6379 --ulimit memlock=-1 docker.dragonflydb.io/dragonflydb/dragonfly 
```

In order to verify the configuration is properly working.

```json
> % cat database-router-config.json 
{
  "*": {
    "name": "redis",
    "options": {
      "url": "redis://localhost:6380"
    }
  }
}
```

`GI_PERSIST=router deno task chel serve dist/` would then hang after this point:

```
{"level":40,"time":1761681093733,"pid":739479,"hostname":"fedora-linux-42","msg":"[SBP WARN]: registering unsafe selector: 'chelonia.db/delete' (remember to lock after overwriting)"}
{"level":40,"time":1761681093734,"pid":739479,"hostname":"fedora-linux-42","msg":"Missing VAPID identification. Please set VAPID_EMAIL to a value like \"mailto:some@example\"."}
{"level":30,"time":1761681093778,"pid":739479,"hostname":"fedora-linux-42","msg":"[database-router] Reading config from path ./database-router-config.json"}
[SBP WARN]: registering unsafe selector: 'chelonia.db/get' (remember to lock after overwriting)
[SBP WARN]: registering unsafe selector: 'chelonia.db/set' (remember to lock after overwriting)
[SBP WARN]: registering unsafe selector: 'chelonia.db/delete' (remember to lock after overwriting)
Missing VAPID identification. Please set VAPID_EMAIL to a value like "mailto:some@example".
{"level":30,"time":1761681093803,"pid":739479,"hostname":"fedora-linux-42","msg":"[database-router] Reading config from path ./database-router-config.json"}
[SBP WARN]: registering unsafe selector: 'chelonia.db/get' (remember to lock after overwriting)
[SBP WARN]: registering unsafe selector: 'chelonia.db/set' (remember to lock after overwriting)
[SBP WARN]: registering unsafe selector: 'chelonia.db/delete' (remember to lock after overwriting)
Missing VAPID identification. Please set VAPID_EMAIL to a value like "mailto:some@example".
(node:739479) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
{"level":30,"time":1761681093815,"pid":739479,"hostname":"fedora-linux-42","msg":"[database-router] Reading config from path ./database-router-config.json"}
(node:739479) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(node:739479) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
```

So I'm not convinced that the config actually is being respected.

PR #85, opened 2025-10-30
Correct changes

### LINKED ISSUE #83: Fix remaining problematic differences between chel serve and GI backend
#### Problem in `routes.ts`

In `routes.ts`:

```md
   - **Change 5**: In `/dev-file` and `/file` POST `failAction`, added `Boom.isBoom(err) ? err : Boom.boomify(err || new Error())`. This ensures non-Boom errors are wrapped, standardizing responses.
     - **Impact**: Low-medium (better error uniformity).
     - **Relevant diff**:
       ```
       +        return Boom.isBoom(err) ? err : Boom.boomify(err || new Error())
       -        return err
       ```
```

This `|| new Error()` pattern is odd and wrong. First, why would it be undefined, and even if it were, the error message should say something, like "there's an unknown error in failAction". Or simply `new Error(err)`.

#### Question for `pubsub.js`

There's now this logic added:

```md
   - **Change 1**: In ping interval, added `if (client.endpoint) return` to skip pinging clients with endpoints (e.g., push subs?).
```

Assuming:

```ts
      (server.clients as Set<unknown> as Set<WS>).forEach((client) => {
        if ((client as unknown as PushSubscriptionInfo).endpoint) return
```

Verify this is correct.

#### Another `pubsub.ts` question

```diff
-          socket.server.emit(error, 'error in socket connection')
+          socket.server.emit('error', error)
```

I could see how maybe the first parameter needed to be a string, but now the 2nd, which used to be a string, is now an error of type unknown.

Shouldn't this be `error?.message || 'error in socket connection'` or something to that effect?

#### `database.ts` question

```diff
 export const initDB = async ({ skipDbPreloading } = {}) => {
   if (persistence) {
-    const Ctor = (await import(`./database-${persistence}.js`)).default
+    const Ctor = (await import(`./database-${persistence}.ts`)).default
     const { init, readData, writeData, deleteData, close } = new Ctor(options[persistence])
     await init()
```

This `import()` thing is a runtime thing right? Then are we sure this will work with `.ts` files, especially when running via the `chel` command? 🤔 

If this is fine, no problem. If not, there are a few other places that do the same type of change that need to be addressed.

### Solution

Go through the above changes and fix / improve anything as needed.

corrideat:
Closes #83

### DIFF
--- a/src/serve/dashboard/controller/backend.ts
+++ b/src/serve/dashboard/controller/backend.ts
@@ -6,7 +6,7 @@ const languageFileMap = new Map([
   ['ko', 'korean.json']
 ])
 
-export function handleFetchResult <T extends 'json' | 'text' | 'blob'>(type: T): ((r: Response) => ReturnType<Response[T]>) {
+export function handleFetchResult <T extends 'json' | 'text' | 'blob'> (type: T): ((r: Response) => ReturnType<Response[T]>) {
   return function (r: Response) {
     if (!r.ok) throw new Error(`${r.status}: ${r.statusText}`)
     return r[type]() as ReturnType<Response[T]>


src/serve/vapid.ts
```diff
@@ -32,7 +32,6 @@ export const initVapid = async () => {
       const serializedKeyPair = await Promise.all([
         crypto.subtle.exportKey('jwk', keyPair.privateKey),
         crypto.subtle.exportKey('raw', keyPair.publicKey).then((key) =>
-
```
corrideat: Extra whitespace removed

src/serve/routes.ts
```diff
@@ -1010,10 +1009,9 @@ if (isCheloniaDashboard) {
...
           // https://w3c.github.io/ServiceWorker/#service-worker-allowed
           if (request.path.includes('assets/js/sw-')) {
-            console.debug('adding header: Service-Worker-Allowed /')
-            if (request.response instanceof Boom.Boom) {
-              request.response.output.headers['Service-Worker-Allowed'] = '/'
-            } else {
+            // If `response` is an error, then no SW is being served
+            if (!(request.response instanceof Boom.Boom)) {
+              console.debug('adding header: Service-Worker-Allowed /')
               request.response.header('Service-Worker-Allowed', '/')
```
corrideat: Based on the discussion on Slack, add the header only for non-error responses (which also is how it works on GI)

src/serve/dashboard/controller/backend.ts
```diff
@@ -6,7 +6,9 @@ const languageFileMap = new Map([
   ['ko', 'korean.json']
 ])
 
-export function handleFetchResult (type: string): ((r: Response) => Promise<unknown>) {
+export function handleFetchResult <T extends 'json' | 'text' | 'blob'>(type: T): ((r: Response) => ReturnType<Response[T]>);
+
+export function handleFetchResult (type: 'json' | 'text' | 'blob'): ((r: Response) => Promise<unknown>) {
```
taoeffect: Sonnet isn't happy with this and says:

There are **several issues**, though it may appear to work:

## 1. **`ReturnType<Response[T]>` syntax issue**

While this *might* work in some TypeScript versions, it's problematic:

```ts
ReturnType<Response[T]>  // T is 'json' | 'text' | 'blob'
```

When `T` is a union type, `Response[T]` gives you the union of all those method types, which isn't what you want. You want the return type to depend on **which specific** value of `T` was passed.

## 2. **The overload doesn't match the implementation**

The implementation signature should be compatible with all overloads, but the generic overload returns a more specific type than `Promise<unknown>`.

## 3. **Dangerous type assertion**

This assertion bypasses all type safety:
```ts
(r as unknown as { [key: string]: () => Promise<unknown> })[type]()
```

---

## Better Approach

```ts
// Function overloads for each specific case
export function handleFetchResult(type: 'json'): (r: Response) => Promise<any>;
export function handleFetchResult(type: 'text'): (r: Response) => Promise<string>;
export function handleFetchResult(type: 'blob'): (r: Response) => Promise<Blob>;

// Implementation
export function handleFetchResult(type: 'json' | 'text' | 'blob') {
  return function (r: Response) {
    if (!r.ok) throw new Error(`${r.status}: ${r.statusText}`);
    return r[type]();  // ✅ No type assertion needed!
  }
}
```

This is **clearer, safer**, and TypeScript can properly infer the return type based on which overload is called.

taoeffect (changes requested):
Nice! One minor concern that Sonnet has, and if it's right I think I agree its implementation looks cleaner:

PR #87, opened 2025-10-31
79 implement chel migrate

### LINKED ISSUE #79: Implement `chel migrate`

corrideat:
Closes #79

### DIFF
--- a/src/migrate.ts
+++ b/src/migrate.ts
@@ -49,6 +49,7 @@ export async function migrate (args: string[]): Promise<void> {
     ['SIGUSR2', 11]
   ] as [string, number][]).forEach(([signal, code]) => handleSignal(signal, code))
 
+  let lastReportedPercentage = 0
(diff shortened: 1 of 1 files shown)


corrideat:
Skip existing keys not implemented because:

1. We don't have a way of checking for a key's existence (we could add a selector for that, however). So, we currently need to read all data.
2. Even if we implemented the previous point, it may or may not be more performant (because of the overhead of the extra lookup to see if the key exists). It will likely be more performant if writing a big chunk of data.

PR #91, opened 2025-11-24
Feature/chel dx

corrideat:
Follow up from #64

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -28412,7 +28412,8 @@ var init_db_utils = __esm({
               // Check if the value is at the end
               currentIndex.endsWith("\0" + value) || // Check if the value is at the start
               currentIndex.startsWith(value + "\0") || // Check if the current index is exactly the value
-              currentIndex === value
+              currentIndex === value || // Check if the value is in the middle
+              currentIndex.includes("\0" + value + "\0")
             ) {
               return;
             }
--- a/build/serve/ownerSizeTotalWorker.js
+++ b/build/serve/ownerSizeTotalWorker.js
@@ -4395,7 +4395,8 @@ var appendToIndexFactory = (key) => {
           // Check if the value is at the end
           currentIndex.endsWith("\0" + value) || // Check if the value is at the start
           currentIndex.startsWith(value + "\0") || // Check if the current index is exactly the value
-          currentIndex === value
+          currentIndex === value || // Check if the value is in the middle
+          currentIndex.includes("\0" + value + "\0")
         ) {
           return;
         }
--- a/src/serve/db-utils.ts
+++ b/src/serve/db-utils.ts
@@ -61,7 +61,9 @@ export const appendToIndexFactory = (key: string): (value: string) => Promise<vo
           // Check if the value is at the start
           currentIndex.startsWith(value + '\x00') ||
           // Check if the current index is exactly the value
-          currentIndex === value
+          currentIndex === value ||
+          // Check if the value is in the middle
+          currentIndex.includes('\x00' + value + '\x00')
         ) {
           // Exit if the value already exists
           return


src/pin.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    if (sourceContent === targetContent) {
+      console.log(colors.gray(`⏭️  Skipping: ${fileName} (unchanged)`))
+      return
+    } else {
+      console.log(colors.blue(`📄 Copying: ${fileName} (changed)`))
+      await copyFile(sourcePath, targetPath)
+      return
+    }
+  }
```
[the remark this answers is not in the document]
taoeffect: That's what `--overwrite` should be used for. Otherwise nothing that exists should be overwritten.

taoeffect (changes requested):
Opus 4.5 review ready!

<details>


## Critical Issues

### 1. Bug: Worker Path Construction Error in `src/serve/server.ts`

The worker path construction appears to be incorrect:

```typescript
: createWorker(join(import.meta.dirname || '.', 'serve', 'ownerSizeTotalWorker.js'))
```

If `import.meta.dirname` points to `src/serve` (the directory containing `server.ts`), then `join(...)` would produce `src/serve/serve/ownerSizeTotalWorker.js` which is wrong.

**Suggested fix:**
```typescript
: createWorker(join(import.meta.dirname || '.', 'ownerSizeTotalWorker.js'))
```

Or if the workers are in a different location after build:
```typescript
: createWorker(new URL('./ownerSizeTotalWorker.js', import.meta.url).toString())
```

---

### 2. Bug: Commented Out Cleanup Code in `scripts/build.ts`

```typescript
  } finally {
-    Deno.removeSync(tmpFile)
+    // Deno.removeSync(tmpFile)
  }
```

This appears to be debugging code accidentally left in. Temporary files will accumulate and not be cleaned up.

**Suggested fix:**
```typescript
  } finally {
    Deno.removeSync(tmpFile)
  }
```

---

### 3. Bug: Incorrect Default Value for `dir` Option in `src/pin.ts`

```typescript
.option('dir', {
  default: false,
  describe: 'Output directory',
  requiresArg: false,
  string: true
})
```

The `default: false` is incorrect for a string option. This will cause type confusion when accessing `args['dir']`.

**Suggested fix:**
```typescript
.option('dir', {
  describe: 'Output directory',
  requiresArg: true,
  string: true
})
```

---

### 4. Bug: `Promise.try` May Not Be Available in `src/serve/createWorker.ts`

```typescript
Promise.try(() => sbp(...(deserializer(msg.data)) as [string, ...unknown[]])).then((r) => {
```

`Promise.try` is a Stage 2+ proposal and may not be available in all JavaScript environments.

**Suggested fix:**
```typescript
Promise.resolve().then(() => sbp(...(deserializer(msg.data)) as [string, ...unknown[]])).then((r) => {
```

Or wrap in an IIFE:
```typescript
(async () => {
  try {
    const r = await sbp(...(deserializer(msg.data)) as [string, ...unknown[]])
    const { data, transferables } = serializer(r)
    port.postMessage([true, data], transferables)
  } catch (e) {
    const { data, transferables } = serializer(e)
    port.postMessage([false, data], transferables)
  } finally {
    port.close()
  }
})()
```

---

## Security Issues

### 5. Potential Path Traversal in `src/pin.ts`

The `sanitizeContractName` function has a logic issue:

```typescript
function sanitizeContractName (contractName: string): string {
  return contractName.replace(/[/\\:*?"<>|]/g, '_').replace(/\.\./g, '__')
}
```

The order of operations means that `foo/../bar` becomes `foo_.__bar` (after `/` replacement) and then the `..` check no longer matches. While this specific case isn't exploitable, the function should be more defensive.

**Suggested fix:**
```typescript
function sanitizeContractName (contractName: string): string {
  // First replace path separators
  let sanitized = contractName.replace(/[/\\:*?"<>|]/g, '_')
  // Then collapse any remaining dots that could form traversal patterns
  sanitized = sanitized.replace(/\.{2,}/g, '_')
  // Ensure it doesn't start or end with dots
  sanitized = sanitized.replace(/^\.+|\.+$/g, '_')
  return sanitized
}
```

---

### 6. Symlink Following in `src/utils.ts`

The `findManifestFiles` function follows symlinks which could lead outside the intended directory:

```typescript
const realPath = await Deno.realPath(join(path, entry.name))
const info = await Deno.lstat(realPath)
if (info.isDirectory) {
  const subitems = await internal(realPath)
```

**Suggested fix - Add an option to not follow symlinks:**
```typescript
export const findManifestFiles = async (path: string, followSymlinks = false): Promise<Set<string>> => {
  const visited = new Set<string>()
  const internal = async (currentPath: string) => {
    const realPath = await Deno.realPath(currentPath)
    if (visited.has(realPath)) {
      return new Set<string>()
    }
    visited.add(realPath)

    const manifests = new Set<string>()
    for await (const entry of Deno.readDir(currentPath)) {
      const entryPath = join(currentPath, entry.name)
      const info = await Deno.lstat(entryPath)
      
      // Skip symlinks unless explicitly following them
      if (info.isSymlink && !followSymlinks) {
        continue
      }
      
      const targetPath = info.isSymlink ? await Deno.realPath(entryPath) : entryPath
      const targetInfo = info.isSymlink ? await Deno.lstat(targetPath) : info
      
      if (targetInfo.isDirectory) {
        const subitems = await internal(targetPath)
        for (const item of subitems) {
          manifests.add(item)
        }
      } else if (entry.name.toLowerCase().endsWith('.manifest.json')) {
        manifests.add(entryPath)
      }
    }

    return manifests
  }

  const realPath = await Deno.realPath(path)
  return internal(realPath)
}
```

---

## Code Quality Issues

### 7. Silent Error Swallowing in `src/deploy.ts`

```typescript
} catch {
  console.warn(`Skipping invalid path: ${manifestPath}`)
  continue
}
```

This hides the actual error that occurred, making debugging difficult.

**Suggested fix:**
```typescript
} catch (error) {
  console.warn(`Skipping invalid path: ${manifestPath}`, error instanceof Error ? error.message : error)
  continue
}
```

---

### 8. Unnecessary Early Return in `src/pin.ts`

```typescript
if (!manifestPath) {
  await loadCheloniaConfig()
  return
}
```

This loads the config and immediately returns without doing anything useful with it.

**Suggested fix:**
```typescript
if (!manifestPath) {
  console.log(colors.yellow('No manifest path provided. Displaying current configuration...'))
  await loadCheloniaConfig()
  console.log(colors.cyan('Current pinned contracts:'))
  for (const [name, info] of Object.entries(cheloniaConfig.contracts)) {
    console.log(colors.gray(`  ${name}: ${info.version} (${info.path})`))
  }
  return
}
```

---

### 9. Missing Await in Watch Function `src/serve.ts`

```typescript
deploy({
  ...args,
  manifests
}).catch((e) => {
  console.warn(e, 'Error deploying contracts')
})
```

The deploy call inside the queue event is not awaited, which means exceptions won't be properly handled by the queue.

**Suggested fix:**
```typescript
sbp('okTurtles.eventQueue/queueEvent', queueName, async () => {
  const manifests = Array.from(manifestSet)
  manifestSet.clear()
  try {
    await deploy({
      ...args,
      manifests
    })
  } catch (e) {
    console.warn(e, 'Error deploying contracts')
  }
})
```

---

### 10. Missing Error Handling in `findManifestFiles`

The function doesn't handle permission errors or other filesystem issues:

```typescript
for await (const entry of entries) {
  const realPath = await Deno.realPath(join(path, entry.name))
```

**Suggested fix:**
```typescript
for await (const entry of entries) {
  try {
    const realPath = await Deno.realPath(join(path, entry.name))
    const info = await Deno.lstat(realPath)
    // ... rest of logic
  } catch (error) {
    // Log warning but continue processing other entries
    console.warn(`Warning: Could not process ${join(path, entry.name)}:`, error instanceof Error ? error.message : error)
    continue
  }
}
```

---

### 11. Inconsistent Visited Set Check in `src/utils.ts`

```typescript
const internal = async (path: string) => {
  const entries = Deno.readDir(path)
  if (visited.has(path)) {
    return new Set<string>()
  }
  visited.add(path)
```

The check happens after `Deno.readDir` is called (though it's lazy, so no I/O yet). Also, `entries` is never awaited or used correctly here - it should be iterated immediately or stored differently.

**Suggested fix:**
```typescript
const internal = async (currentPath: string) => {
  if (visited.has(currentPath)) {
    return new Set<string>()
  }
  visited.add(currentPath)
  
  const manifests = new Set<string>()
  for await (const entry of Deno.readDir(currentPath)) {
    // ... rest of logic
  }
  return manifests
}
```

---

### 12. Removed IPC Handler May Break Subprocess Communication in `src/serve/index.ts`

The following was removed:
```typescript
-process.on('message', (message) => {
-  console.info('message received in child, shutting down...', message)
-  exit(0)
-})
```

If this server is spawned as a child process (e.g., by grunt as mentioned in the TODO comment), removing this handler will break graceful shutdown via IPC.

**Consider keeping this or explicitly documenting the removal:**
```typescript
// NOTE: IPC message handling removed - if running as subprocess,
// use signals (SIGTERM) for graceful shutdown instead
```

---

## Minor Issues

### 13. Type Annotation in `src/serve/database.ts`

```typescript
value = prefixHandlers[prefix](value) as string | Buffer
```

The type assertion could mask type errors. Consider proper typing of `prefixHandlers`.

---

### 14. Inconsistent Alias in `src/serve.ts`

The dashboard-port alias changed from `'d'` to `'a'`, while `'d'` is now used for `dev`:

```typescript
-.alias('d', 'dashboard-port')
+.alias('a', 'dashboard-port')
+.alias('d', 'dev')
```

This is a breaking change for existing users who use `-d` for dashboard port. Consider using a different alias for `dev` to maintain backward compatibility.

</details>

README.md
```diff
@@ -1,11 +1,16 @@
...
+Modern CLI for Chelonia contract development, deployment, and management.
+
+## All Available Commands
 
 ```
 chel
 chel help [command]
 chel version
+chel build
+chel pin <version> <manifest-file-path> [--only-changed | --overwrite]
```
taoeffect: - `--only-changed` should be removed here in the readme as well
- The `<version> should come at the end
- The options should be before the mandatory params

src/serve.ts
```diff
@@ -61,8 +132,22 @@ export const module = {
         requiresArg: true,
         number: true
       })
-      .alias('d', 'dashboard-port')
+      .alias('a', 'dashboard-port')
```
taoeffect: Didn't catch this - should be using `-a` for `dashboard-port`

PR #109, opened 2025-12-19
Fix data corruption

### LINKED ISSUE #108: `chel migrate` results in corrupted files
### Problem

> So changing the URL did work! It seems to be loading the site fine now from Redis! Only issue is that some attachments don't work, like videos:
> 
> ```
> AttachmentDownloadItem.vue caught: Error: mismatched chunk size
> ```

Investigating this showed that this doesn't affect _just_ redis, but all migrations. This happens because `chelonia.db/get` will convert buffers to strings, using UTF-8 encoding, and this transformation corrupts binary data.

### Solution

Use the `any:` prefix to avoid data corruption.

corrideat:
Closes #108

### DIFF
--- a/src/migrate.ts
+++ b/src/migrate.ts
@@ -40,6 +40,7 @@ export async function migrate (args: ArgumentsCamelCase<Params>): Promise<void> 
     await backendTo.init()
   } catch (error) {
     exit(error)
+    throw error
   }
 
   const numKeys = await sbp('chelonia.db/keyCount')
@@ -100,7 +101,7 @@ export async function migrate (args: ArgumentsCamelCase<Params>): Promise<void> 
     }
     // `any:` prefix needed to get the raw value, else the default is getting
     // a string, which will be encoded as UTF-8. This can cause data loss.
-    let value: Buffer | string
+    let value: Buffer | string | undefined
     try {
       value = await sbp('chelonia.db/get', `any:${key}`)
     } catch (e) {
@@ -108,6 +109,7 @@ export async function migrate (args: ArgumentsCamelCase<Params>): Promise<void> 
       console.error(`Error reading from source database key '${key}'`, e)
       await backendTo.close()
       exit(1)
+      throw e
     }
     await checkAndExit()
     // Make `deno check` happy.
@@ -122,6 +124,7 @@ export async function migrate (args: ArgumentsCamelCase<Params>): Promise<void> 
       console.error(`Error writing to target database key '${key}'`, e)
       await backendTo.close()
       exit(1)
+      throw e
     }
     await checkAndExit()
     ++numMigratedKeys
--- a/build/main.js
+++ b/build/main.js
@@ -109555,6 +109555,7 @@ async function migrate(args) {
     await backendTo.init();
   } catch (error) {
     exit(error);
+    throw error;
   }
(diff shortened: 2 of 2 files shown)


src/migrate.ts
```diff
@@ -42,55 +42,94 @@ export async function migrate (args: ArgumentsCamelCase<Params>): Promise<void>
...
-  const handleSignal = (signal: string, code: number) => {
-    process.on(signal, () => {
-      console.error(`Received signal ${signal} (${code}). Finishing current operation.`)
-      // Exit codes follow the 128 + signal code convention.
-      // See <https://tldp.org/LDP/abs/html/exitcodes.html>
-      shouldExit = 128 + code
-    })
+
+  const reportStatus = () => {
+    console.log(`[chel] ${colors.green('Migrated:')} ${numMigratedKeys} entries`)
```
taoeffect: Why are various log statements here prefixed with `[chel]`?

Don't we know that this is the chel command being run?

src/migrate.ts
```diff
@@ -42,55 +42,94 @@ export async function migrate (args: ArgumentsCamelCase<Params>): Promise<void>
...
+      continue
+    }
+    // `any:` prefix needed to get the raw value, else the default is getting
+    // a string, which will be encoded as UTF-8. This can cause data loss.
+    let value: unknown
+    try {
+      value = await sbp('chelonia.db/get', `any:${key}`)
+    } catch (e) {
+      reportStatus()
+      console.error('Error reading from source databse', key, e)
```
taoeffect: There's a typo here and below in `databse`.

Also, what about making the key a bit clearer in the log message?

E.g. something like:

```ts
console.error(`Error reading from source database key '${key}':`, e)
```

src/migrate.ts
corrideat: Not sure, that's how it was before

taoeffect:
Hopefully final run of Opus:

<details>


### Resource leak on error paths

When an error occurs during read or write operations, `exit(1)` is called without closing `backendTo`:

```typescript
    } catch (e) {
      reportStatus()
      console.error('Error reading from source databse', key, e)
      exit(1)
    }
```

and

```typescript
    } catch (e) {
      reportStatus()
      console.error('Error writing to target databse', key, e)
      exit(1)
    }
```

**Suggested fix:**
```typescript
    } catch (e) {
      reportStatus()
      console.error('Error reading from source database', key, e)
      await backendTo.close()
      exit(1)
    }
```

```typescript
    } catch (e) {
      reportStatus()
      console.error('Error writing to target database', key, e)
      await backendTo.close()
      exit(1)
    }
```

---

### Division by zero when database is empty

If the source database has no keys (`numKeys === 0`), this will result in `Infinity`:

```typescript
    const percentage = Math.floor((numVisitedKeys / numKeys) * 100)
```

**Suggested fix:**
```typescript
    const percentage = numKeys > 0 ? Math.floor((numVisitedKeys / numKeys) * 100) : 100
```

Alternatively, add an early check at the start of the migration:

```typescript
  const numKeys = await sbp('chelonia.db/keyCount')
  if (numKeys === 0) {
    console.log('[chel] Source database is empty. Nothing to migrate.')
    return
  }
```

</details>

Issue #105, opened 2025-12-18
chel.toml validation


taoeffect:
### Problem

The `chel.toml` file currently isn't validated very well in some parts, and for other parts (like issue #104), the error messages for existing validation aren't very helpful.

### Solution

- Provide validation for the chel.toml file in the sense that if parameters are specified, they contain correct values
- If keys are valid but the values are invalid (e.g. `localhost: 6379` for `url` instead of `redis://localhost: 6379`), provide helpful error messages like `"url" must begin with redis://, rediss://, or unix://`)
- If keys are invalid (don't exist), a warning or an error should be printed. It might be worth printing an error if mandatory keys that are expected aren't defined, and a warning for other keys that are unknown (as this could indicate a typo)
- The `uncaught (in promise)` issue as noted in #104 should be fixed.

Issue #104, opened 2025-12-16
Trouble running `chel serve`


taoeffect:
### Problem

After seemingly performing a successful migration from sqlite to redis using `chel migrate`, I tried running `grunt dev` on the GI `chel serve` PR and got:

```
-> % DENO_DIR=$PWD/.deno grunt dev  
Running "exec:gitconfig" (exec) task

Running "checkDependencies:this" (checkDependencies) task

Running "chelDeploy" task

Running "exec:chelDevDeploy" (exec) task
>> [SBP WARN]: registering unsafe selector: 'chelonia.db/get' (remember to lock after overwriting)
>> [SBP WARN]: registering unsafe selector: 'chelonia.db/set' (remember to lock after overwriting)
>> [SBP WARN]: registering unsafe selector: 'chelonia.db/delete' (remember to lock after overwriting)
>> [SBP WARN]: registering unsafe selector: 'chelonia.db/iterKeys' (remember to lock after overwriting)
>> [SBP WARN]: registering unsafe selector: 'chelonia.db/keyCount' (remember to lock after overwriting)
>> error: Uncaught (in promise) TypeError: Invalid protocol
>>           throw new TypeError("Invalid protocol");
>>                 ^
>>     at RedisClient.parseURL (file:///tmp/deno-compile-chel/build/main.js:44431:17)
>>     at RedisClient.parseOptions (file:///tmp/deno-compile-chel/build/main.js:44412:30)
>>     at Class2.#initiateOptions (file:///tmp/deno-compile-chel/build/main.js:44595:37)
>>     at new RedisClient (file:///tmp/deno-compile-chel/build/main.js:44520:46)
>>     at new Class2 (file:///tmp/deno-compile-chel/build/main.js:43009:49)
>>     at file:///tmp/deno-compile-chel/build/main.js:44404:32
>>     at create (file:///tmp/deno-compile-chel/build/main.js:44408:37)
>>     at createClient3 (file:///tmp/deno-compile-chel/build/main.js:54058:40)
>>     at RedisBackend.init (file:///tmp/deno-compile-chel/build/main.js:54118:55)
>>     at initDB (file:///tmp/deno-compile-chel/build/main.js:54588:17)
>> Exited with code: 123.
>> Error executing child process: Error: Process exited with code 123.
Warning: Task "exec:chelDevDeploy" failed. Use --force to continue.

Aborted due to warnings.
```

So then I tried running raw `chel serve`, but got the same result:

```
[...]

PR #111, opened 2025-12-20
Add DB schema validation

corrideat:
(EDIT: @taoeffect): ~~Closes~~ Almost #105

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -71,33 +71,33 @@ var deno_default = {
 var isWindows = globalThis.Deno?.build.os === "windows" || globalThis.navigator?.platform?.startsWith("Win") || globalThis.process?.platform?.startsWith("win") || false;
 
 // deno:https://jsr.io/@std/path/1.0.9/_common/from_file_url.ts
-function assertArg3(url) {
-  url = url instanceof URL ? url : new URL(url);
-  if (url.protocol !== "file:") {
-    throw new TypeError(`URL must be a file URL: received "${url.protocol}"`);
+function assertArg3(url2) {
+  url2 = url2 instanceof URL ? url2 : new URL(url2);
+  if (url2.protocol !== "file:") {
+    throw new TypeError(`URL must be a file URL: received "${url2.protocol}"`);
   }
-  return url;
+  return url2;
 }
 
 // deno:https://jsr.io/@std/path/1.0.9/posix/from_file_url.ts
-function fromFileUrl(url) {
-  url = assertArg3(url);
-  return decodeURIComponent(url.pathname.replace(/%(?![0-9A-Fa-f]{2})/g, "%25"));
+function fromFileUrl(url2) {
+  url2 = assertArg3(url2);
+  return decodeURIComponent(url2.pathname.replace(/%(?![0-9A-Fa-f]{2})/g, "%25"));
 }
 
 // deno:https://jsr.io/@std/path/1.0.9/windows/from_file_url.ts
-function fromFileUrl2(url) {
-  url = assertArg3(url);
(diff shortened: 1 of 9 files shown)


taoeffect (commented):
Nice! I liked how you went about addressing this!

Does this PR fix this issue btw?

>  The `uncaught (in promise)` issue as noted in #104 should be fixed.

Here's the Opus 4.5 review:

<details>


I found several issues with the Zod schema definitions and a minor issue in the migrate error handling.

## Critical: Invalid Zod API Usage in `database-redis.ts`

```typescript
const ConfigSchema = z.strictObject({
  url: z.optional(z.string('').url({ protocol: /^rediss?$/ })),
})
```

There are two problems here:
1. `z.string()` doesn't accept a string argument
2. `z.string().url()` doesn't accept a `protocol` option in Zod

**Suggested fix:**

```typescript
const ConfigSchema = z.strictObject({
  url: z.optional(
    z.string().url().refine(
      (url) => /^rediss?:\/\//.test(url),
      { message: 'URL must use redis:// or rediss:// protocol' }
    )
  ),
})
```

## Critical: Invalid Zod API Usage in `database-router.ts`

```typescript
const ConfigEntrySchema = z.strictObject({
  name: z.string(),
  options: z.object()
})
```

`z.object()` requires a shape argument—it cannot be called without arguments. Based on the TypeScript type `Record<string, unknown>`, this should allow arbitrary keys.

**Suggested fix:**

```typescript
const ConfigEntrySchema = z.strictObject({
  name: z.string(),
  options: z.record(z.string(), z.unknown())
})
```

## Bug: Schema Stricter Than TypeScript Type in `database-router.ts`

```typescript
const ConfigSchema = z.intersection(
  z.object({ '*': ConfigEntrySchema }),
  z.record(z.string(), ConfigEntrySchema)
)
```

This schema *requires* the `'*'` key to be present, but the TypeScript type and constructor default allow an empty object:

```typescript
constructor (config: Config = {}) {
```

Passing `{}` would satisfy the TypeScript type but fail Zod validation.

**Suggested fix** (if `'*'` should be optional):

```typescript
const ConfigSchema = z.record(z.string(), ConfigEntrySchema)
```

Or if `'*'` is truly required, update the TypeScript type to match and remove the default empty object.

## Minor: Potentially Unreachable Code in `migrate.ts`

```typescript
  } catch (e) {
    console.error('Error setting up database')
    exit(e)
    throw e
  }
```

If `exit(e)` terminates the process (e.g., via `process.exit()`), the `throw e` is unreachable dead code. If `exit` is asynchronous or doesn't actually terminate execution, the `throw` is needed for control flow.

**Suggestion:** If `exit` does terminate, consider adding a `// @ts-expect-error` or `return` comment to clarify intent, or use TypeScript's `never` return type on `exit` so the compiler knows control doesn't continue. If `throw e` is intentional for type-narrowing purposes, a brief comment would help future readers.

</details>

taoeffect (changes requested):
Great work @corrideat ! I like the solution you've taken. I did some testing:

Is `chel serve` supposed to fail if I do this?

```toml
zbackend = "router"
```

Note the `z` prefix. I see no errors or warnings, it seems to default to `mem` when this happens.

Note this condition of #105:

> If keys are invalid (don't exist), a warning or an error should be printed. It might be worth printing an error if mandatory keys that are expected aren't defined, and a warning for other keys that are unknown (as this could indicate a typo)

taoeffect (approved):
Alright, this seems like it partially addresses #105.

To close #105 yes we should manually load the TOML file and verify that its config passes via a whitelist of config options that could be there.

PR #126, opened 2026-03-19
Fix manifest version

corrideat:
This fixes an important issue with `chel manifest`. Apparently, `yargs` (unless this is disabled) will add a `--version / v` option. This is fine, but it doesn't allow it to be overridden by commands. The result is that `chel manifest` will always consider the version as unspecified.

---

<a href="https://app.devin.ai/review/okturtles/chel/pull/126" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ chel version
 chel pin [--dir <output-directory>] [--overwrite] <manifest-file-path> <version>
 chel test
 chel keygen [--out=<key.json>]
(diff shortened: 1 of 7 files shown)


taoeffect (approved):
LGTM!

PR #129, opened 2026-04-02
Report version from manifest

corrideat:
These changes read version information from `chelonia.json` instead of from environment variables and also present per-contract versions (as opposed to a single, global, `CONTRACTS_VERSION`)

### DIFF
--- a/src/pin.ts
+++ b/src/pin.ts
@@ -6,13 +6,15 @@ import process from 'node:process'
 import type { ArgumentsCamelCase, CommandModule } from './commands.ts'
 import { exit } from './utils.ts'
 
+const RESERVED_FILE_CHARS = /[/\\:*?"<>|]/g
+
 type Params = { overwrite: boolean, 'dir'?: string, 'manifest-version'?: string, manifest: string }
 
 let projectRoot: string
 let cheloniaConfig: { contracts: Record<string, { version: string, path: string }> }
 
 function sanitizeContractName (contractName: string): string {
-  return contractName.replace(/[/\\:*?"<>|]/g, '_').replace(/\.\./g, '__')
+  return contractName.replace(RESERVED_FILE_CHARS, '_').replace(/\.\./g, '__')
 }
 
 export async function pin (args: ArgumentsCamelCase<Params>): Promise<void> {
@@ -37,6 +39,11 @@ export async function pin (args: ArgumentsCamelCase<Params>): Promise<void> {
     }
 
     const { contractName, fullContractName, contractFiles, manifestVersion } = await parseManifest(fullManifestPath)
+
+    if (RESERVED_FILE_CHARS.test(manifestVersion)) {
+      exit(`Invalid manifest version: ${manifestVersion}`)
+    }
+
     console.log(colors.blue(`Contract name: ${fullContractName}`))
     console.log(colors.blue(`Manifest version: ${manifestVersion}`))
 
--- a/src/pin.ts
+++ b/src/pin.ts
@@ -6,7 +6,7 @@ import process from 'node:process'
 import type { ArgumentsCamelCase, CommandModule } from './commands.ts'
 import { exit } from './utils.ts'
 
-const RESERVED_FILE_CHARS = /[/\\:*?"<>|]/g
+const RESERVED_FILE_CHARS = /[/\\:*?"<>|]/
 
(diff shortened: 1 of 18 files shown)


corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/23916491640)._

src/pin.ts
```diff
@@ -50,31 +50,31 @@ export async function pin (args: ArgumentsCamelCase<Params>): Promise<void> {
...
       }
-      console.log(colors.yellow(`Version ${manifestVersion} already exists for ${contractName} - checking files...`))
+      console.log(colors.yellow(`Version ${manifestVersion} already exists for ${fullContractName} - checking files...`))
     } else {
       await createVersionDirectory(contractName, manifestVersion)
     }
 
     await copyContractFiles(contractFiles, manifestPath, contractName, manifestVersion, args)
-    await updateCheloniaConfig(contractName, manifestVersion, manifestPath)
+    await updateCheloniaConfig(fullContractName, manifestVersion, manifestPath)
```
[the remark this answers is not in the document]
corrideat: This change is intentional to keep the name in `chelonia.json`.

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24487427890)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24490484000)._

taoeffect (changes requested):
Looks solid, I think! Just have a documentation request:

Where is `chelonia.json` documented?

Need to know:

- What it is
- What it's for
- What it can contain
- How it's different from the .toml config

Could you add this to the README.md maybe?

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24611521653)._

taoeffect (approved):
LGTM! 👍

src/serve/server.ts
```diff
@@ -435,8 +446,11 @@ sbp('okTurtles.data/set', PUBSUB_INSTANCE, createServer(hapi.listener, {
   serverHandlers: {
     connection (socket) {
       const versionInfo = {
-        GI_VERSION: GI_VERSION || null,
-        CONTRACTS_VERSION: CONTRACTS_VERSION || null
+        appVersion: cheloniaAppManifest?.appVersion || null,
+        contractsVersion: cheloniaAppManifest?.contracts ? Object.fromEntries(
+          Object.entries(cheloniaAppManifest?.contracts)
+            .map(([k, v]) => [k, (v as { version: string }).version])
+        ) : null
```
taoeffect, line 453: Question: does this design assume there will only be one app per server?

If so, that's clearly not a safe assumption (per the other open issues).

So I think we should address this somehow in this PR in a way that either fully addresses it, or sets up future versions to address it properly.

I.e.:

- It could fully address this by returning all apps
- It could partially address this by passing in some server API version perhaps that clients could distinguish for figuring out what format to expect?

taoeffect (commented):
Oh, actually, I missed one thing:

src/serve/server.ts
corrideat: _At the moment_ there can only possibly be one app per server. Since this is being sent when a WS connection is established, I think this is fine (since presumably, a WS client will be for a single app).

I think it's low risk to make this assumption since, whichever changes are needed, they are limited to the server API and will not require changing contracts.

In addition, presumably, we'll want `https://groupincome.app` (as an example) to be a single app, regardless of how many apps the backend is running. I.e., the origin will identify the app.

With this in mind, these changes look safe and perhaps even future proof:

1. We might need to refactor `chel` to serve multiple apps, but those changes don't seem related to the ones here (it'd mean, in this case, that we'd have to pick which `cheloniaAppManifest` we want --- now, there's a single choice)
2. In the future, since we need to pick a way of choosing the `cheloniaAppManifest` (see previous point), we need a way to do it. If we do it based on origin, this already is future proof. If we do it some other way, we can change the WS protocol to add an app id selector (which would not require recreating contracts)

corrideat: > Question: does this design assume there will only be one app per server?

Currently, yes. But, I think little if anything will change from the client's perspective.

PR #131, opened 2026-04-14
Replace hapi with hono

### LINKED ISSUE #69: 'Failed to fetch' issues when running the Cypress tests
## Problem

It seems like when running Cypress tests and the app server is using a Deno backend, sometimes 'Failed to fetch' errors occur, causing the tests to fail.

I've tried looking into the cause for this in the past and wasn't able to find it. However, I've only seen it happening on Cypress, and it can be 'fixed' by putting an nginx reverse-proxy in front of the Deno backend (so that Cypress doesn't access Deno directly).

Based on what happens, I'm guessing that the issue is in the network packets being sent. It could be something in Cypress (unlikely, since I believe it uses the same network stack as Chrome / Chromium) or in Deno, or specifically in the combination of Deno and Hapi.

This issue happens sporadically, so it may not be very high priority.

## Solution

Investigate what causes it and fix it if possible.

### LINKED ISSUE #75: Move off of Hapi? (CodeQL issue + perf)
### Problem

[CodeQL is complaining about the built file](https://github.com/okTurtles/chel/runs/53610822704). This apparently has to do with Hapi JS, something that's somewhat out of our control.

### Solution

If Hapi doesn't fix this soon, we could move off of Hapi.js to something that's more Deno specific and possibly more performant too.

corrideat:
This closes #75 and closes #69


taoeffect:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24377471620)._

src/serve/auth.ts
```diff
@@ -1,52 +1,68 @@
...
+  return { token: authorization.slice(prefix.length) }
+}
 
-    server.auth.strategy('chel-bearer', 'chel-bearer')
-    server.auth.strategy('chel-shelter', 'chel-shelter')
+function extractShelter (c: Context): AuthCredentials | null {
+  const authorization = c.req.header('authorization')
+  if (!authorization) return null
+  const prefix = 'shelter '
+  if (authorization.slice(0, prefix.length) !== prefix) return null
```
corrideat: This needs .toLowerCase() on the LHS

src/serve/auth.ts
```diff
@@ -1,52 +1,68 @@
...
+    credentials: AuthCredentials
+    authStrategy: string
+  }
+}
+
+function extractBearer (c: Context): AuthCredentials | null {
+  const authorization = c.req.header('authorization')
+  if (!authorization) return null
+  const prefix = 'bearer '
+  if (authorization.slice(0, prefix.length) !== prefix) return null
```
corrideat: This needs .toLowerCase() on the LHS

src/serve/routes.ts
corrideat, line 1: Changes here mostly seem fine so far (need to go more thoroughly). However, I'm noticing that schema validation (declarative) has been replaced by code. This isn't wrong necessarily, but makes it harder to understand the code.

taoeffect: Yeah that needs to be fixed and restored to using a declarative style. Clearly Opus didn't get the Zod message

src/pin.ts
```diff
@@ -50,31 +50,31 @@ export async function pin (args: ArgumentsCamelCase<Params>): Promise<void> {
...
       }
-      console.log(colors.yellow(`Version ${manifestVersion} already exists for ${contractName} - checking files...`))
+      console.log(colors.yellow(`Version ${manifestVersion} already exists for ${fullContractName} - checking files...`))
     } else {
       await createVersionDirectory(contractName, manifestVersion)
     }
 
     await copyContractFiles(contractFiles, manifestPath, contractName, manifestVersion, args)
-    await updateCheloniaConfig(contractName, manifestVersion, manifestPath)
+    await updateCheloniaConfig(fullContractName, manifestVersion, manifestPath)
```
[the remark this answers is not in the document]
corrideat: This is intended. The name is only used in `chelonia.json`.

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24587340442)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24589053958)._

taoeffect:
@corrideat here's the review (ran locally with code-review skill):

<details>

# Code Review

**Base**: `master`
**Head**: `pr` (5eb98e1)
**Date**: 2026-04-17

---

This PR replaces the Hapi HTTP framework with Hono across the server implementation, refactors the server lifecycle into explicit `startServer`/`stopServer` functions, replaces Joi validation with Zod, replaces Boom errors with `HTTPException`, and adds comprehensive endpoint tests.

## 1. 🔴 `getConnInfo` imported from `hono/deno` won't work with `@hono/node-server` adaptor

- [x] Addressed
- [ ] Dismissed

`routes.ts:8` imports `getConnInfo` from `npm:hono/deno`, but the HTTP server is created via `createAdaptorServer` from `npm:@hono/node-server` (`server.ts:432`). The Deno-specific `getConnInfo` is designed for `Deno.serve()`'s native request objects, not Node.js compatibility `http.Server` requests. This means the `getConnInfo()` call in `getClientIP()` (`routes.ts:270`) will likely throw or return undefined data in production, causing the function to always fall back to `'unknown'` when there's no `x-real-ip` header.

**Impact**: Without a reverse proxy setting `x-real-ip`, all clients appear as `'unknown'`, effectively sharing a single rate-limit bucket. This defeats the per-IP signup rate limiting (`routes.ts:396-406`).

**Suggestion**: Use `getConnInfo` from `npm:hono/node-server` instead, or rely solely on the `x-real-ip` / `x-forwarded-for` headers (which is the documented requirement anyway):

```typescript
// routes.ts — replace the import
- import { getConnInfo } from 'npm:hono/deno'
+ import { getConnInfo } from 'npm:@hono/node-server'

// Or, remove getConnInfo entirely and use header-based detection:
function getClientIP (c: Context): string {
  return c.req.header('x-real-ip')
    || c.req.header('x-forwarded-for')?.split(',')[0].trim()
    || 'unknown'
}
```

## 2. 🔴 `stopServer()` / `closeDB()` not guarded against concurrent invocation

- [x] Addressed
- [ ] Dismissed

Both `server.ts:631` `stopServer()` and `database.ts:338` `closeDB()` are async functions that mutate module-scoped state (`currentHttpServer`, `currentBackend`, `currentCache`, etc.) but have no re-entrancy guard. They can be invoked concurrently from:

1. Signal handler → `exit()` → `SERVER_EXITING` event → `index.ts` handler calls `stopServer()`
2. `backend/server/stop` SBP selector → `stopServer()`
3. A second signal (e.g., `SIGINT` followed quickly by `SIGTERM`)

This can cause double-close errors, `TypeError: Cannot read properties of null`, or incomplete cleanup.

**Suggestion**: Add a guard flag:

```typescript
// server.ts
let isStopping = false

export async function stopServer (): Promise<void> {
  if (isStopping) return
  isStopping = true
  try {
    // ... existing cleanup ...
  } finally {
    isStopping = false
  }
}
```

Similarly for `closeDB()` in `database.ts`.

## 3. 🟡 Rate limiter cleanup accesses Bottleneck private `interval` property

- [ ] Addressed
- [x] Dismissed

`routes.ts:321` accesses `(limiter as unknown as { interval: ReturnType<typeof setInterval> }).interval` to clean up an internal timer that Bottleneck v2's `Group.disconnect()` doesn't clear. This relies on an undocumented private property that could break on any Bottleneck minor/patch update.

The comment documents the workaround, which is good, but this is still fragile. Consider wrapping this in a helper with a version check, or pinning the Bottleneck version explicitly in `deno.json`.

## 4. 🟡 Three catch blocks rely on `errorMapper`'s `: never` return type with no fallback

- [x] Addressed
- [ ] Dismissed

At `routes.ts:514`, `routes.ts:859`, and `routes.ts:917`, `errorMapper(err)` is the last statement in a `catch` block with no explicit `throw` after it. The pattern works only because `errorMapper` is typed `: never` and always throws. However, if `errorMapper` is ever refactored to have a non-throwing branch (or if a new error type is added that falls through the `switch`), these catch blocks would silently return `undefined`, resulting in a `200 OK` with an empty body — a confusing response for an error.

Other catch blocks in the same file explicitly `throw err` after logging, which is more robust.

**Suggestion**: Add an explicit `throw` after `errorMapper` calls, or change `errorMapper` to return the exception instead of throwing it, so callers must explicitly throw:

```typescript
// Option A — explicit throw (minimal change)
catch (err) {
  // ... logging ...
  errorMapper(err as Error)
  throw err // unreachable, but defensive
}

// Option B — errorMapper returns the exception
const errorMapper = (e: Error): HTTPException => {
  switch (e?.name) {
    case 'BackendErrorNotFound': return new HTTPException(404)
    // ...
    default: return new HTTPException(500, { message: e.message ?? 'internal error' })
  }
}
// Caller: throw errorMapper(err as Error)
```

## 5. 🟡 Dashboard server's `Deno.serve` has no shutdown mechanism

- [ ] Addressed
- [x] Dismissed

`dashboard-server.ts:44` calls `Deno.serve()` but never stores the returned server handle. The `startDashboard()` function returns `Promise<void>`, so there's no way for callers to stop the dashboard server during shutdown. This means:

- On graceful shutdown, the dashboard port remains open while the main server closes
- Tests that start/stop the server can't clean up the dashboard
- In development with `--dev` watch mode, old dashboard servers may leak

**Suggestion**: Return the server handle so it can be closed:

```typescript
export async function startDashboard (): Promise<Deno.HttpServer> {
  // ...
  const server = Deno.serve({ port, hostname: host, onListen: () => resolve() }, app.fetch)
  return server
}
```

## 6. ⚪️ `getStaticServeConfig()` returns stale Hapi-style route path

- [x] Addressed
- [ ] Dismissed

`routes.ts:237-244` — `getStaticServeConfig()` returns `routePath` with Hapi syntax (`'/dashboard/{path*}'` or `'/app/{path*}'`), but this value is never used in the Hono route registration. The Hono routes use hardcoded paths like `'/assets/:subpath{.+}'` and `'/dashboard/*'`. The `routePath` and `redirect` fields are used (for the root redirect), but `routePath` itself is dead. Minor cleanup opportunity.

## 7. ⚪️ `genericWorker.ts` initializes logger at import time before nconf is ready

- [ ] Addressed
- [x] Dismissed

`genericWorker.ts:5` calls `initializeLogger()` at module scope. Worker processes import this module, but `nconf` configuration may not be loaded yet at that point (the logger reads `LOG_LEVEL` from nconf at `logger.ts:32`). Since `initializeLogger()` has an idempotency guard, a second call from the main server would be a no-op, meaning workers might get the default log level rather than the configured one. This is pre-existing behavior (the old code imported `./logger.ts` at module scope too), so this is informational rather than a regression.

</details>

corrideat:
One remaining issue: remote address shows up as `unknown` when there's no `x-real-ip`. This might be a Hono bug.

src/serve/routes.ts
```diff
@@ -146,1042 +217,949 @@ const ctEq = (expected: string, actual: string): boolean => {
...
-          // header to allow the service worker to function. Details:
-          // https://w3c.github.io/ServiceWorker/#service-worker-allowed
-          if (request.path.includes('assets/js/sw-')) {
-            // If `response` is an error, then no SW is being served
-            if (!(request.response instanceof Boom.Boom)) {
-              console.debug('adding header: Service-Worker-Allowed /')
-              request.response.header('Service-Worker-Allowed', '/')
-            }
+          if (result) {
+            return c.json(result)
```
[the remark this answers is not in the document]
corrideat: Fixed

scripts/build.ts
```diff
@@ -48,10 +49,23 @@ if (result.errors.length) {
...
+import { builtinModules } from 'node:module'
+
+const nodeBuiltins = new Set(builtinModules.filter((m: string) => !m.startsWith('_')))
+const bareBuiltinRe = /\bfrom\s+(["'])([^"']+)\1/g
+
+function addNodePrefix (source: string): string {
+  return source.replace(bareBuiltinRe, (match, _, specifier) => {
+    return nodeBuiltins.has(specifier) ? `from "node:${specifier}"` : match
+  })
+}
```
corrideat: Perhaps this could be done properly with an esbuild plugin re-mapping the import, instead of using a brittle text replace.

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24678585200)._

corrideat:
<details>
<summary>
Opus 4.7 (via Copilot)
</summary>

# Code Review

- **Base**: `master`
- **Head**: `replace-hapi-with-hono` (6b1317d)
- **Date**: 2026-04-20
- **Scope**: Replaces Hapi + Joi + Boom with Hono + Zod + HTTPException; introduces async start/stop server lifecycle. 24 non-build files changed across server, routes, auth, config, CLI bootstrapping, and tests.

Overall the migration is coherent and the Hono-based routing is cleaner. The main risks are lifecycle/shutdown semantics (SERVER_EXITING no longer emitted globally, selectors no longer locked) and a handful of small behavioral regressions compared to the Joi/Hapi originals.

---

## 1. 🔴 `SERVER_EXITING` is no longer emitted after a command completes

- [ ] Addressed
- [x] Dismissed

`src/main.ts` used to wrap `postHandler()` in `try { ... } finally { sbp('okTurtles.events/emit', SERVER_EXITING) }`. The new code drops the `try/finally` entirely and just `await`s `handlerState.postHandler()`.

Consequences:
- Any listener registered on `SERVER_EXITING` (e.g. external cleanup, log flushing) will never fire on normal command completion.
- Cleanup has been moved piecemeal into individual commands: `src/get.ts`, `src/upload.ts`, `src/eventsAfter.ts`, and `src/migrate.ts` now each call `closeDB()` in their own `finally`. Commands that don't do this (e.g. `hash`, `keygen`, `manifest`, `deploy`, `pin`, `verifySignature`, `latestState`) no longer participate in the global shutdown event at all.

If this was intentional, add a comment in `main.ts` explaining the new contract (each command owns its own teardown). Otherwise, restore the emit:

```ts
try {
  await handlerState.postHandler()
} finally {
  sbp('okTurtles.events/emit', SERVER_EXITING)
}
```

## 2. 🔴 `database.ts` no longer locks its SBP selectors

- [x] Addressed
- [ ] Dismissed

The previous implementation finished `initDB` with `sbp('sbp/selectors/lock', [...])` for the `chelonia.db/*` and `backend/db/*` selectors. The new `src/serve/database.ts` removes those lock calls entirely.

This weakens the defense-in-depth guarantee that contract code (or any other module) cannot silently replace the database layer. If the reason for removal is to allow `closeDB()` + re-`initDB()` cycles (useful in tests and in `stopServer`/`startServer`), that's a legitimate trade-off — but please either:
- add a comment documenting why the selectors are intentionally left unlocked, or
- re-lock them at the end of `initDB`, and unlock them in `closeDB` before any replacement.

## 3. 🔴 `X-Frame-Options` middleware sets headers after `await next()` for streamed responses

- [ ] Addressed
- [x] Dismissed

In `src/serve/server.ts` (≈ lines 420–430), the middleware does:

```ts
app.use('*', async (c, next) => {
  await next()
  c.header('X-Frame-Options', 'DENY')
})
```

For routes that construct a `Response` directly with a streaming body (notably `/eventsAfter/*` in `routes.ts`, which returns `new Response(Readable.toWeb(stream), { headers })`), the headers on that `Response` are already committed by the time this middleware resumes. `c.header(...)` after `next()` only modifies `c.res`, and whether Hono re-applies it to a user-constructed `Response` depends on Hono's internals — in practice this header is frequently lost for streamed replies.

The old Hapi `onPreResponse` hook ran before the response was flushed and was reliable. Verify by curling `/eventsAfter/...` and checking the response headers; if missing, merge the header into the `Response` construction explicitly or use `c.res = new Response(c.res.body, { ...c.res, headers: ... })` pattern.

## 4. 🟡 `parseBooleanParam` does not match its own Joi-compat comment

- [x] Addressed
- [ ] Dismissed

`src/serve/routes.ts` (≈ 24–32) documents itself as "Matches `Joi.boolean().truthy('1','on','yes').falsy('0','off','no')`", but:
- It returns `true` for empty string (so `?keyOps=` becomes `true`). Joi's default would coerce this to `false` or reject.
- It returns `false` for any unrecognised value silently, whereas Joi would reject with a 400.

Either tighten the function to throw on unknown values (and let `errorMapper` turn it into a 400), or fix the comment to describe what it actually does. Suggested stricter version:

```ts
const TRUTHY = new Set(['1', 'on', 'yes', 'true'])
const FALSY  = new Set(['0', 'off', 'no', 'false', ''])
function parseBooleanParam (v: string | undefined, field: string): boolean {
  if (v === undefined) return false
  const s = v.toLowerCase()
  if (TRUTHY.has(s)) return true
  if (FALSY.has(s)) return false
  throw new HTTPException(400, { message: `Invalid boolean for ${field}` })
}
```

## 5. 🟡 `zValidatorFormOrJson` silently JSON-parses when Content-Type is missing/unexpected

- [x] Addressed
- [ ] Dismissed

`src/serve/routes.ts` (≈ 119–149): if the request's Content-Type is not `application/x-www-form-urlencoded`, the validator falls through to `c.req.json()` regardless of whether Content-Type is `application/json`, missing, or something else entirely. The error surface ("Invalid request body") is also less informative than Joi's original per-field messages.

Consider:
- Requiring one of the two Content-Types explicitly and returning 415 otherwise.
- Including Zod's `error.issues` in the 400 response body (see `errorMapper`).

Also, for `/zkpp/register/:name` the Zod schema is `z.union([{ b }, { r, s, sig, Eh }])`. A payload like `{ b, r, s }` (extra fields from the second branch) will match the first branch and the extras are silently dropped. The old Joi schema would have rejected. Consider `.strict()` on each branch.

## 6. 🟡 `parseArgs()` now eagerly `.parse()`s and is then passed to `nconf.argv()`

- [x] Addressed
- [ ] Dismissed

`src/parseArgs.ts` calls `yargsInstance.parse()` at the end and returns the instance. `src/parseConfig.ts:12` then does `.argv(parseArgs())`. `nconf.argv(...)` typically calls `.argv` on the yargs instance itself, which can cause yargs to re-run its command pipeline.

Confirm that the command's `builder`/`handler` is not executed twice (add a temporary `console.log` in one handler and run any command). If it is, switch to either:
- passing `parseArgs().parsed.argv` (the already-parsed result) to nconf, or
- not calling `.parse()` in `parseArgs` and letting nconf trigger it.

## 7. 🟡 `migrate.ts` uses `backendTo!.close()` in `finally` despite possible `undefined`

- [x] Addressed
- [ ] Dismissed

`src/migrate.ts` (≈ line 155):

```ts
} finally {
  await Promise.all([backendTo!.close(), closeDB()])
}
```

If `backendTo` throws during `init()` before assignment (or if the `try` throws before it's constructed), the non-null assertion is a lie. In practice `exit(error)` is called first and short-circuits via `Deno.exit(1)`, but that relies on non-local reasoning. Prefer:

```ts
await Promise.all([backendTo?.close(), closeDB()])
```

## 8. 🟡 `/eventsAfter` streaming: a DB error mid-stream produces a truncated 200

- [x] Addressed
- [ ] Dismissed

`src/serve/routes.ts` (≈ 515) reads `streamHeaders` before the stream is consumed and wraps the Node stream via `Readable.toWeb(stream)`. If the underlying DB read errors partway through, the HTTP status is already 200, the body aborts mid-flight, and the client has no structured way to distinguish this from a complete response.

This is arguably no worse than the Hapi implementation, but now would be a good time to at least log the stream error server-side:

```ts
stream.on('error', (err) => logger.error('eventsAfter stream error', err))
```

…and document the semantics in a comment.

## 9. 🟡 Rate-limiter cleanup on re-`startServer()` is correct but subtle

- [ ] Addressed
- [ ] Dismissed

`installRateLimiterSelectorsOnce()` in `routes.ts` (~234) and `installServerSelectorsOnce()` in `server.ts` use module-level "once" flags while the state they operate on (`currentLimiters`, `currentApp`, `currentBackend`) is module-let. On a second `startServer()` in the same process, `registerRoutes` first tears down the previous limiters (`routes.ts:330`), then builds new ones. The selector `backend/server/stopRateLimiters` re-reads `currentLimiters` at call-time, so it stays correct.

Not a bug, but:
- Add a short comment at the "once" installation sites explaining the invariant ("selectors close over the module-level mutable slot, which is overwritten in start/stop").
- Consider registering these selectors unconditionally on module load (they're idempotent closures) and removing the "once" flag + boolean.

## 10. 🟡 Dev-only `POST /log` assumes JSON without checking Content-Type

- [ ] Addressed
- [x] Dismissed

`src/serve/routes.ts` (≈ 549–559) calls `c.req.json()` unconditionally. A non-JSON body throws an unhandled `SyntaxError` that surfaces as a 500 via `errorMapper`. Minor, since this route only exists in development, but wrapping with the standard `zValidatorFormOrJson` (or at least `try/catch`) would be more consistent with the rest of the file.

## 11. ⚪ Backwards-compat `app` proxy exposes `undefined` before `startServer()`

- [x] Addressed
- [ ] Dismissed

`src/serve/server.ts` (≈ 701–708) exports a `Proxy` that forwards property access to `currentApp`. Before `startServer()` runs, `currentApp` is null and `app.fetch` (etc.) is `undefined` — silently, not a thrown error. The declared type is `Hono`, which is now a lie.

Either:
- Remove the export and migrate callers to `startServer()`/explicit access, or
- Throw in the proxy's `get` trap when `currentApp === null`:

```ts
get (_t, prop) {
  if (!currentApp) throw new Error('Server not started; call startServer() first')
  return Reflect.get(currentApp, prop, currentApp)
}
```

## 12. ⚪ Dashboard SPA fallback passes a noop `next` to `serveStatic`

- [ ] Addressed
- [x] Dismissed

`src/serve/dashboard-server.ts` (≈ 33–42) invokes the `serveStatic` middleware manually with `async () => {}` as `next` and inspects the returned `Response`. This relies on Hono's `serveStatic` returning a `Response` when a file is found and calling `next()` (without returning one) when it isn't — an implementation detail that's not part of Hono's public contract.

A more idiomatic approach is to register `serveStatic` as middleware on the sub-app and then register a final catch-all `app.get('*', ...)` that serves `index.html`. That way the SPA fallback is just a normal route.

## 13. ⚪ `safePathWithin` is correct but should have direct tests

- [ ] Addressed
- [x] Dismissed

`src/serve/routes.ts` (≈ 293–297) correctly resolves the target path and compares against `base + path.sep` to prevent `/foo` matching `/foobar`. Given that this is the critical guard on file-serving endpoints, add explicit unit tests in `routes-test-helpers.ts` / a new `routes-static.test.ts` for:

- `../../etc/passwd`
- URL-encoded traversal (`%2e%2e%2f...`)
- Null-byte injection
- Absolute paths
- Paths whose resolved form equals `base` (should be allowed) vs `basefoo` (should not).

## 14. ⚪ Chunk-collection loop stops at first gap

- [ ] Addressed
- [x] Dismissed

`src/serve/routes.ts` (≈ 720–725): `for (let i = 0; ; i++) { if (!chunkFile) break }` silently truncates at the first missing chunk index. If a client posts chunks `0, 1, 3`, the server collects `[0, 1]` and may then fail schema validation with a confusing "missing chunk" error, or — worse — accept the upload missing chunk 2.

Either accept only contiguous chunks and validate count against an expected header, or explicitly detect gaps and return a 400 with a clear message.

---

## Summary

- **Critical (🔴)**: 3 — missing `SERVER_EXITING` emit, unlocked DB selectors, `X-Frame-Options` on streamed responses.
- **Medium (🟡)**: 7 — boolean parsing, body validation loosening, yargs/nconf double-parse risk, `migrate.ts` non-null assertion, streaming error handling, rate-limiter install comments, dev `/log` body parsing.
- **Minor (⚪)**: 4 — backwards-compat `app` proxy, dashboard SPA fallback pattern, missing `safePathWithin` tests, chunk-gap behavior.

No blockers for the migration itself; the Hono port is generally faithful. Recommended before merge: fix #1, #2, and #3; address #4 and #6 to avoid subtle behavioral regressions versus the Joi/Hapi baseline.

</details>

src/serve/routes.ts
```diff
@@ -37,49 +52,101 @@ const KV_KEY_REGEX = /^(?!_private)[^\x00]{1,256}$/
...
+  z.object({ b: z.string() }),
+  z.object({ r: z.string(), s: z.string(), sig: z.string(), Eh: z.string() })
+])
+const zkppUpdatePasswordBodySchema = z.object({
+  r: z.string().min(1, 'r is required'),
+  s: z.string().min(1, 's is required'),
+  sig: z.string().min(1, 'sig is required'),
+  hc: z.string().min(1, 'hc is required'),
+  Ea: z.string().min(1, 'Ea is required')
+})
```
taoeffect: Is there really no way to keep the same type of inline structure that we had before with Joi? Do we have to bunch all of these validations up here instead of doing them inline where they're used?

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/24689129519)._

taoeffect:
Could also real quick benchmark the current server vs this one to verify that in fact Hono is faster than Hapi?

GI has a backend unit test that posts events. You could benchmark posting and reading events.

src/serve/routes.ts
```diff
@@ -389,799 +378,746 @@ if (process.env.NODE_ENV === 'development') {
...
-    const cid = existing ? createCID(existing, multicodes.RAW) : ''
 
-    if (expectedEtag === '*') {
-    // pass through
-    } else {
-      if (!expectedEtag.split(',').map((v: string) => v.trim()).includes(`"${cid}"`)) {
-        // We need this `x-cid` because HAPI modifies Etags
-        return h.response(existing || '').etag(cid).header('x-cid', `"${cid}"`).code(412)
+    const payloadBuffer = Buffer.from(await c.req.arrayBuffer())
+
```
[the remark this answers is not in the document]
corrideat: This isn't really a regression, but the issue is valid. See #133.

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/25013376925)._

taoeffect:
/crush_fast {{openrouter, deepseek/deepseek-v4-pro}}

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/25015789346)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/25017304806)._

src/serve/routes.ts
```diff
@@ -161,1042 +186,998 @@ const ctEq = (expected: string, actual: string): boolean => {
...
+        const chunks: [string, Uint8Array][] = await Promise.all(manifest.chunks.map(async (chunk: [number, string], i: number) => {
+        // Validate the chunk information
+          if (
+            !Array.isArray(chunk) ||
+          chunk.length !== 2 ||
+          typeof chunk[0] !== 'number' ||
+          typeof chunk[1] !== 'string' ||
+          !Number.isSafeInteger(chunk[0]) ||
+          chunk[0] <= 0
+          ) {
```
taoeffect, line 707: The indentation here is off. Lines 702-706 should be indented. Please check elsewhere in the code for similar indentation problems, or ask `deepseek/deepseek-v4-flash` to do it for you (free on Venice).

corrideat: Good catch. Fixed this, as well as ran prettier to fix other issues (apparently, there's a conflict with our `eslint` rules that interfere with the `prettier` plugin). It mostly fixed line length issues and inconsistent quotes / semi-colon used. I'm also reviewing the diff now and I asked a few LLMs to identify indentation issues.

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/25260119775)._

corrideat:
One preexisting but important issue is this:

> Issue 5 — `getClientIP` trusts X-Real-IP header without validation, enabling IP spoofing for rate limiting bypass

If the server is directly exposed to the Internet, for whatever reason, limits are trivial to circumvent by setting to `x-real-ip` to any arbitrary, client-chosen value.

taoeffect:
GPT-5.5:

<details>

## 1. 🟡 Security headers are skipped for CORS preflights and thrown errors

- [x] Addressed
- [ ] Dismissed

The Hapi implementation added `X-Frame-Options: DENY` from an `onPreResponse` extension for both normal and Boom error responses. In the Hono version, the header is installed as middleware after the CORS middleware in `src/serve/server.ts:419-425`:

```ts
currentApp.use('*', cors({ origin: '*' }))

currentApp.use('*', async (c, next) => {
  await next()
  c.header('X-Frame-Options', 'DENY')
})
```

Because Hono's CORS middleware handles `OPTIONS` preflight requests without calling later middleware, those responses never get the header. Also, if a route throws an `HTTPException` or other error, `await next()` rejects and the header-setting line is skipped. This is a regression from the previous Hapi `onPreResponse` path that handled Boom responses explicitly.

Move the header before CORS and make error responses set it too, for example:

```ts
currentApp.use('*', async (c, next) => {
  await next()
  c.header('X-Frame-Options', 'DENY')
})
currentApp.use('*', cors({ origin: '*' }))
currentApp.onError((err, c) => {
  const response = err instanceof HTTPException
    ? err.getResponse()
    : c.text('Internal Server Error', 500)
  response.headers.set('X-Frame-Options', 'DENY')
  return response
})
```

## 2. 🟡 Missing dashboard assets return `index.html` with status 200

- [ ] Addressed
- [x] Dismissed

The dashboard fallback in `src/serve/dashboard-server.ts:35-43` tries static serving and then returns `index.html` for every unmatched path:

```ts
app.get('/*', async (c) => {
  const staticResponse = await staticMiddleware(c, async () => {})
  if (staticResponse) {
    return staticResponse
  }
  const indexResponse = await indexMiddleware(c, async () => {})
  return indexResponse || c.text('Not Found', 404)
})
```

This means a missing script, stylesheet, or image such as `/assets/missing.js` is served as the SPA HTML with `200 OK`. Browsers will then fail with confusing MIME/type errors and caches may store the wrong response. The old Hapi `/assets/{path*}` static route would return a real 404 for missing files instead of falling through to the SPA.

Only use the SPA fallback for navigation requests or extensionless paths, and return 404 for asset-like paths:

```ts
app.get('/*', async (c) => {
  const staticResponse = await staticMiddleware(c, async () => {})
  if (staticResponse) return staticResponse

  const requestedPath = new URL(c.req.url).pathname
  const acceptsHtml = c.req.header('accept')?.includes('text/html')
  if (path.extname(requestedPath) && !acceptsHtml) {
    return c.text('Not Found', 404)
  }

  return await indexMiddleware(c, async () => {}) || c.text('Not Found', 404)
})
```

## 3. ⚪️ Bare `/app` and `/dashboard` application routes no longer match

- [x] Addressed
- [ ] Dismissed

The application SPA route now registers only `'/app/*'` or `'/dashboard/*'` in `src/serve/routes.ts:1054`. Hono wildcard routes do not cover the bare prefix without a trailing slash, so direct links to `/app` or `/dashboard` fall through instead of serving or redirecting to the SPA. Hapi's `/{path*}` style route accepted the empty trailing path.

Register the bare prefix explicitly next to the wildcard route:

```ts
const spaPrefix = isCheloniaDashboard ? '/dashboard' : '/app'
app.get(spaPrefix, (c) => c.redirect(`${spaPrefix}/`))
app.get(`${spaPrefix}/*`, etag(), async function (c) {
  const file = await Deno.readFile(staticServeConfig.distIndexHtml)
  return c.body(file, 200, { 'Content-Type': 'text/html' })
})
```

## 4. ⚪️ Reinitializing the DB with the in-memory backend after a persistent backend leaves stale selectors

- [ ] Addressed
- [x] Dismissed

The new explicit DB lifecycle closes a persistent backend by setting `currentBackend = null` in `src/serve/database.ts:317-327`, but `initDB()` only overwrites the `chelonia.db/*` selectors inside the `persistence && persistence !== 'mem'` branch in `src/serve/database.ts:217-285`. If a long-lived process or test suite initializes a persistent backend, calls `closeDB()`, and later initializes with `database.backend = 'mem'`, the persistent selectors remain installed and call `currentBackend!.readData(...)` while `currentBackend` is `null`.

Either restore the default in-memory selectors on close, or overwrite selectors for the `mem` path as well. At minimum, reset `setSelectors` when closing an unlocked persistent backend so the next non-mem init cannot reuse stale closures:

```ts
if (!(import.meta as ImportMeta).lockDbSelectors) {
  setSelectors = false
  preloaded = false
  currentCache = null
}
```

This should be paired with an explicit mem-selector restore if switching back to `mem` in the same process is supported.

</details>

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/25338913406)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/25342117976)._

taoeffect (approved):
Great work @corrideat! 👏 👏 👏

PR #140, opened 2026-06-20
Add ETag / x-cid to POST /kv response

corrideat:
This is part of the KV refactoring in `@chelonia/lib`. Now, we return the `etag` / `x-cid` of a KV write.

Before merging this PR, it'd make sense to have a new `@chelonia/lib` release with the KV changes.

This PR does _not_ depend on the changes being made to `@chelonia/lib`, but we also don't currently have a use for the `etag` / `x-cid` headers in 204 responses, which we will once we merge https://github.com/okTurtles/libcheloniajs/pull/80 and https://github.com/okTurtles/group-income/pull/3133.


taoeffect:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/27981528309)._

taoeffect (dismissed):
LGTM, but will wait on GLM

taoeffect:
Ricardo says that we just need to do it consistently, and that by default we return strings when retrieving values from the database, so we should use strings here for computing the ETag as well.

So the main issue is basically that we store data as bytes by default, and retrieve it as strings by default.

So to prevent this confusion, at the very least we should add a comment to the codebase acknowledging this and explain why we're doing the hash on the string as opposed to the bytes, and then also verify that we do this consistently everywhere else we create a CID.

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/28375107441)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/28377652275)._

src/serve/routes.ts
```diff
@@ -938,7 +938,11 @@ export function registerRoutes (app: Hono): void {
       // Use a queue to prevent race conditions (for example, writing to a contract
       // that's being deleted or updated)
       return sbp('chelonia/queueInvocation', contractID, async () => {
-        const existing = await sbp('chelonia.db/get', `_private_kv_${contractID}_${key}`)
+        // Read with the `any:` prefix so the value comes back as the raw stored
+        // bytes (Buffer) rather than being UTF-8 decoded to a string. This makes
+        // the ETag a true content-address of what was persisted, regardless of
+        // whether the bytes are valid UTF-8.
+        const existing = await sbp('chelonia.db/get', `any:_private_kv_${contractID}_${key}`)
```
corrideat: Wrap in `Buffer.from()` for safety (since `any:` _could_ return a string)

src/serve/routes.ts
```diff
@@ -1008,7 +1022,10 @@ export function registerRoutes (app: Hono): void {
         throw new HTTPException(401)
       }
 
-      const result = await sbp('chelonia.db/get', `_private_kv_${contractID}_${key}`)
+      // Read with the `any:` prefix to get the raw stored bytes (Buffer) instead
+      // of a UTF-8 decoded string, so the ETag content-addresses exactly what was
+      // persisted and matches the CID the POST/412/409 paths return.
+      const result = await sbp('chelonia.db/get', `any:_private_kv_${contractID}_${key}`)
```
corrideat: Same here (wrap in `Buffer.from`)

corrideat:
LGTM

taoeffect (approved):
LGTM!

PR #143, opened 2026-06-22
Fix `JSON.parse` issues on `deleteFile`

### LINKED ISSUE #142: `Error parsing manifest`  at `deleteFile`
### Problem

When running `chel serve`, I've observed messages like the following in the logs: 

```
[18:02:54.143] WARN (13993): Error parsing manifest for zLGQo2DimCH8PQx1Wnh1tqxdmkCTyVEStuizjs7wpiwknnSUg2QV4ywp. It's probably not a file manifest.
    err: {
      "type": "SyntaxError",
      "message": "Unexpected non-whitespace character after JSON at position 3 (line 1 column 4)",
      "stack":
          SyntaxError: Unexpected non-whitespace character after JSON at position 3 (line 1 column 4)
              at JSON.parse (<anonymous>)
              at Object.backend/deleteFile (file:///tmp/user/1000/deno-compile-chel/build/main.js:76152:32)
              at async PersistentAction.attempt (file:///tmp/user/1000/deno-compile-chel/build/main.js:71051:22)
    }
```

In this case, `zLGQo2DimCH8PQx1Wnh1tqxdmkCTyVEStuizjs7wpiwknnSUg2QV4ywp` _is_ a valid file manifest (which can be verified using the `/file/zLGQo2DimCH8PQx1Wnh1tqxdmkCTyVEStuizjs7wpiwknnSUg2QV4ywp` route) _and_ therefore also is valid JSON.

The likely cause is that the data being accessed are in a `Buffer` / `Uint8Array` object instead of a string, which trips `JSON.parse`.

### Solution

Fix JSON parsing so that this code works and doesn't result in an error.

corrideat:
AI disclosure: These changes were assisted by GPT-5.5 and Opus 4.8

Closes #142


corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/27975600670)._

corrideat:
<details>
<summary>Review by Opus 4.8</summary>

# Code Review

**Base**: `master`
**Head**: `142-error-parsing-manifest-at-deletefile` (`5bf5275`)
**Date**: 2026-06-22

---

## Summary

The branch fixes issue #142 ("Error parsing manifest at deleteFile"). The root
cause is correctly identified and addressed:

- `backend/deleteFile` (`src/serve/server.ts:228-232`) previously called
  `JSON.parse(rawManifest)` directly. With the in-memory backend,
  `chelonia.db/get` returns the stored value untouched, and the empty (`''`)
  prefix handler in `@chelonia/lib/db` only decodes values for which
  `Buffer.isBuffer(value)` is true. A `Uint8Array` is **not** a `Buffer`, so a
  manifest stored as a `Uint8Array` came back as a `Uint8Array`, and
  `JSON.parse` stringified it to comma-separated bytes (`"123,34,..."`) →
  parse failure → `BackendErrorNotFound`. The fix now converts any non-string
  value via `Buffer.from(rawManifest).toString()` before parsing.
- `POST /file` (`src/serve/routes.ts:750`) now stores the manifest as a string
  (`manifestText`) instead of a `Uint8Array`, so new uploads round-trip cleanly
  through every backend and the legacy decoding path is not needed for new data.

Both behaviors are covered by new tests (`routes-writes.test.ts`): a legacy
byte-manifest deletion path and a full upload-then-delete round trip. All write
tests pass, and the bundled `build/main.js` matches the `src` changes. The three
failing tests in the suite (`routes-reads`, `routes-stateless`, `routes-zkpp`)
fail with pre-existing "message port not closed" leaks unrelated to this branch.

No correctness issues were found that block this change. A few low-severity
observations follow.

## 1. ⚪ Manifest CID is derived from bytes but the stored value is now a string

- [ ] Addressed
- [x] Dismissed

`src/serve/routes.ts:729` still computes `manifestHash` from the original bytes
(`createCID(manifestPayload, ...)`), while line 750 stores the decoded string
(`manifestText`). This is correct in practice — JSON is valid UTF-8, and
`Buffer.from(bytes).toString()` followed by re-encoding is lossless for valid
UTF-8 — so the served `/file` content reproduces the exact bytes the CID was
computed from. Worth a one-line comment noting the intentional bytes-for-hash /
string-for-storage split so a future reader doesn't "fix" it into an
inconsistency:

```ts
// Hash is computed over the raw bytes; the value is stored as text so it
// round-trips identically through every DB backend (see issue #142).
const manifestHash = createCID(manifestPayload, multicodes.SHELTER_FILE_MANIFEST)
```

## 2. ⚪ Chunk re-wrapping with `Buffer.from(data)` is a defensive copy, not strictly required

- [ ] Addressed
- [x] Dismissed

`src/serve/routes.ts:749` changed chunk storage from `data` (a `Uint8Array`) to
`Buffer.from(data)`. `Buffer` is a `Uint8Array` subclass, so serving via the
`any:` route was already functional with the raw `Uint8Array`. The change is
harmless and improves consistency with how other backends hand values back, but
note it allocates a copy per chunk. Not a problem at expected chunk sizes; just
confirming it is intentional rather than load-bearing for the fix.

## 3. ⚪ `deleteFile` still funnels `BackendErrorBadData` into `BackendErrorNotFound`

- [x] Addressed
- [ ] Dismissed

In `src/serve/server.ts:233-240`, the validation throws (`unsupported manifest
version`, `missing chunks`, etc.) are caught by the surrounding `try/catch` and
re-thrown as `BackendErrorNotFound`, masking the more specific
`BackendErrorBadData`. This is pre-existing behavior unchanged by this branch,
but since the file now also handles a new (legacy-bytes) code path, it would be
slightly clearer to scope the `try/catch` to the `JSON.parse` call only, so that
genuine bad-data manifests surface as bad data rather than "not found". Optional.
➜  chel git:(142-error-parsing-manifest-at-deletefile) ✗ xterm
➜  chel git:(142-error-parsing-manifest-at-deletefile) ✗ cat REVIEW.md
# Code Review

**Base**: `master`
**Head**: `142-error-parsing-manifest-at-deletefile` (`5bf5275`)
**Date**: 2026-06-22

---

## Summary

The branch fixes issue #142 ("Error parsing manifest at deleteFile"). The root
cause is correctly identified and addressed:

- `backend/deleteFile` (`src/serve/server.ts:228-232`) previously called
  `JSON.parse(rawManifest)` directly. With the in-memory backend,
  `chelonia.db/get` returns the stored value untouched, and the empty (`''`)
  prefix handler in `@chelonia/lib/db` only decodes values for which
  `Buffer.isBuffer(value)` is true. A `Uint8Array` is **not** a `Buffer`, so a
  manifest stored as a `Uint8Array` came back as a `Uint8Array`, and
  `JSON.parse` stringified it to comma-separated bytes (`"123,34,..."`) →
  parse failure → `BackendErrorNotFound`. The fix now converts any non-string
  value via `Buffer.from(rawManifest).toString()` before parsing.
- `POST /file` (`src/serve/routes.ts:750`) now stores the manifest as a string
  (`manifestText`) instead of a `Uint8Array`, so new uploads round-trip cleanly
  through every backend and the legacy decoding path is not needed for new data.

Both behaviors are covered by new tests (`routes-writes.test.ts`): a legacy
byte-manifest deletion path and a full upload-then-delete round trip. All write
tests pass, and the bundled `build/main.js` matches the `src` changes. The three
failing tests in the suite (`routes-reads`, `routes-stateless`, `routes-zkpp`)
fail with pre-existing "message port not closed" leaks unrelated to this branch.

No correctness issues were found that block this change. A few low-severity
observations follow.

## 1. ⚪ Manifest CID is derived from bytes but the stored value is now a string

- [ ] Addressed
- [ ] Dismissed

`src/serve/routes.ts:729` still computes `manifestHash` from the original bytes
(`createCID(manifestPayload, ...)`), while line 750 stores the decoded string
(`manifestText`). This is correct in practice — JSON is valid UTF-8, and
`Buffer.from(bytes).toString()` followed by re-encoding is lossless for valid
UTF-8 — so the served `/file` content reproduces the exact bytes the CID was
computed from. Worth a one-line comment noting the intentional bytes-for-hash /
string-for-storage split so a future reader doesn't "fix" it into an
inconsistency:

```ts
// Hash is computed over the raw bytes; the value is stored as text so it
// round-trips identically through every DB backend (see issue #142).
const manifestHash = createCID(manifestPayload, multicodes.SHELTER_FILE_MANIFEST)
```

## 2. ⚪ Chunk re-wrapping with `Buffer.from(data)` is a defensive copy, not strictly required

- [ ] Addressed
- [ ] Dismissed

`src/serve/routes.ts:749` changed chunk storage from `data` (a `Uint8Array`) to
`Buffer.from(data)`. `Buffer` is a `Uint8Array` subclass, so serving via the
`any:` route was already functional with the raw `Uint8Array`. The change is
harmless and improves consistency with how other backends hand values back, but
note it allocates a copy per chunk. Not a problem at expected chunk sizes; just
confirming it is intentional rather than load-bearing for the fix.

## 3. ⚪ `deleteFile` still funnels `BackendErrorBadData` into `BackendErrorNotFound`

- [ ] Addressed
- [ ] Dismissed

In `src/serve/server.ts:233-240`, the validation throws (`unsupported manifest
version`, `missing chunks`, etc.) are caught by the surrounding `try/catch` and
re-thrown as `BackendErrorNotFound`, masking the more specific
`BackendErrorBadData`. This is pre-existing behavior unchanged by this branch,
but since the file now also handles a new (legacy-bytes) code path, it would be
slightly clearer to scope the `try/catch` to the `JSON.parse` call only, so that
genuine bad-data manifests surface as bad data rather than "not found". Optional.

</details>

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/28373355649)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/28378311573)._

corrideat:
LGTM

src/serve/server.ts
```diff
@@ -224,17 +224,29 @@ function installServerSelectorsOnce (): void {
...
+          throw new BackendErrorBadData('manifest is not valid JSON')
+        }
         console.warn(e, `Error parsing manifest for ${cid}. It's probably not a file manifest.`)
         throw new BackendErrorNotFound()
       }
+      if (!manifest || typeof manifest !== 'object') throw new BackendErrorBadData('manifest format is invalid')
+      if (manifest.version !== '1.0.0') throw new BackendErrorBadData('unsupported manifest version')
+      if (!Array.isArray(manifest.chunks) || !manifest.chunks.length) throw new BackendErrorBadData('missing chunks')
+      // Delete all chunks
+      await Promise.all(manifest.chunks.map(([, cid]: [unknown, string]) => sbp('chelonia.db/delete', cid)))
```
taoeffect, line 249: Previously this was inside of the try expression and any exceptions would be caught, and a warning logged. This line has now been moved outside of it and also doesn't have a `.catch` attached to it. Double-checking, is that intended? If so, could you add a brief sentence comment on why that's OK to line 248?

corrideat: I don't think this changes the `throw` behaviour in an unsafe way and this looks better:

Before:

1. `Promise.all()` fails
3. Enter `catch`
4. Log `Error parsing manifest for ${cid}. It's probably not a file manifest.`
5. `throw new BackendErrorNotFound()`

An error is still thrown but the message logged is incorrect and the error being thrown is misleading.

Now:

1. `Promise.all()` fails
2. The correct error is thrown instead of `BackendErrorNotFound`

corrideat: (because an error is still being thrown regardless, I don't think this merits a comment)

taoeffect (approved):
Approved!

PR #147, opened 2026-07-17
Implement server ID

### LINKED ISSUE #114: Add 'server ID' config option
### Problem

We currently use variants of the same DB in various environments, such as production and staging. Doing this could result in issues with sending push notifications to unintended clients. At the same time, this setup could be beneficial for testing and debugging purposes.

The unintended notifications can result in issues such as a 'forked chain' error in the client, or less serious issues like unintended contract updates.

### Solution

In order to address this, we could add a `server_id` or similar configuration setting to `chel.toml`, which is stored when a push notification client is stored, and then compared when loading clients from the DB. If a push client has a different `server_id` than what is configurd, it is simply skipped.

In addition, we can create a `chel init` command that generates a template `chel.toml` file to ease set up.

##### router DB engine

The issue can be avoided without any changes by using the `router` DB engine to keep the VAPID keys server-local. This solution is clean and effective, but requires using `router` and requires specific configuration that may or may not apply to some use cases.

##### Automatic `server_id`

Instead of a user-configured `server_id`, it could be automatically obtained from system information, such as the hostname or a machine fingerprint. This solution was discussed but disregarded because of the risk of false negatives (e.g., push notifications would stop working if system configuration changes) and false positives (e.g., the issue could persist when using exact templates to run different servers)

corrideat:
Closes #114

## AI disclosure

Planned and assisted with GLM 5.2.


corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/29592536176)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/29600780073)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/29604536900)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/29606933995)._

corrideat:
LGTM

src/serve/server.ts
```diff
@@ -597,24 +617,72 @@ export async function startServer (): Promise<{ uri: string }> {
...
+          await deleteSubscriptionFromIndex(subscriptionId)
+        } else {
+          skipped++
+        }
+        return
+      }
+      loadIntoMemory(subscriptionId, settings, subscriptionInfo, channelIDs)
     }))
+    if (missing || adopted || reclaimed || skipped) {
+      console.warn(`[server] push-subscription restore: ${adopted} adopted (legacy), ${reclaimed} reclaimed (deleted, server_id mismatch), ${skipped} skipped (server_id mismatch, retained on disk), ${missing} missing (de-indexed)`)
```
taoeffect, line 689: If you're gonna do this "reclamation" thing, users probably don't know how to address it, as they are unlikely to know about `reclaimForeignSubscriptions` as an option they can set.

So if you want to keep it, I recmomend:

1. Before logging it, build this string in a way that it never displays 0 values for any of these variables, and only shows values that are non-zero
2. When building for `reclaimForeignSubscriptions`, print instructions on how to set this variable and what it means to set it.

taoeffect (changes requested):
Nice work @corrideat! One very minor logging update request (see above comment)

src/serve/server.ts
corrideat: Not sure I follow what you mean regarding the second point.

Nevertheless, on '[users] are unlikely to know about reclaimForeignSubscriptions': This is by design. `reclaimForeignSubscriptions` is supposed to be used in exceptional cases, or at least if you fully understand the implications, since it makes destructive changes to the database.

> Before logging it, build this string in a way that it never displays 0 values for any of these variables, and only shows values that are non-zero

This seems like something that will complicate logic for little gain. Having 0 values shown as 0 is useful for grepping and to know that the value truly is 0 and not something else but that the filter is broken.

Can you please clarify what you mean by the following?

> When building for reclaimForeignSubscriptions, print instructions on how to set this variable and what it means to set it

taoeffect: Ah, I see, I think I misunderstood it as "reclaimable", not "reclaimed".

If there's a boolean they need to turn on to reclaim stuff, how are they to know about it? I guess I'm saying is it possible to let them know there are "reclaimable entries"?

See it makes no sense the way it is now, as the default is don't reclaim anything and don't even inform them about it. In such cases it's better to not implement the feature at all. And if you want it, there needs to be some way of telling users about its existence and its applicabIlity in some situation where they'd want to enable it.

corrideat: I see. When it's applicable, I guess a message could be shown about this configuration option.

src/serve/server.ts
```diff
@@ -683,6 +683,9 @@ export async function startServer (): Promise<{ uri: string }> {
     if (missing || adopted || reclaimed || skipped) {
       console.warn(`[server] push-subscription restore: ${adopted} adopted (legacy), ${reclaimed} reclaimed (deleted, server_id mismatch), ${skipped} skipped (server_id mismatch, retained on disk), ${missing} missing (de-indexed)`)
     }
+    if (skipped) {
+      console.warn(`[server] ${skipped} push subscription(s) belong to a different server_id and were left untouched. If this server_id change is intentional and permanent, set 'reclaimForeignSubscriptions = true' under [server] in chel.toml (or set the environment variable server__reclaimForeignSubscriptions=true) to delete these reclaimable entries on next startup. Otherwise, restore the previous server_id to reactivate them.`)
```
taoeffect, line 692: Typo: `belong` => `belonging`

corrideat: I believe this is correct as is.

$X push subscriptions (subject) belong to a different server_id and [the $X push subscriptions] were left untouched.

If you want _belonging_, then I believe we must remove the _and_:

push subscription(s) belonging to a different server_id were left untouched

corrideat:
Approved

taoeffect (approved):
LGTM!

PR #152, opened 2026-07-25
Update `@chelonia/lib`, send `cid` in KV pubsub frames

### LINKED ISSUE #151: Fable: Server never sends `cid` in KV pubsub frames — the self-echo machinery is inert
### Problem

From [a review](https://github.com/okTurtles/libcheloniajs/pull/80#discussion_r3635027070) of [PR #80 in libcheloniajs](https://github.com/okTurtles/libcheloniajs/pull/80), Fable found this issue:

The whole §4.9 design (echo suppression, `fromConflict` markers, last-write-wins gating) keys on the frame's `cid`. The spec asserts "`createKvMessage` now includes the server data `cid` in the frame," but on `chel` master, `routes.ts:1008` calls `broadcastKV(contractID, key, payloadString)` and `server.ts:127` calls `createKvMessage(contractID, key, entry)` — the optional 4th `cid` parameter (`pubsub/index.ts:425`) is never passed. No open chel PR adds it (checked, including #124 "KV attribution").

Against today's server, every frame takes the no-cid path in `_handleRemote` (`kv.ts:2186`), so:

- **Your own write's echo is not suppressed.** It triggers an authoritative GET, a second `CHELONIA_KV_UPDATED` (with `reason: 'remote'`) and a **second `onUpdate` invocation** per local write — the spec explicitly promises self-echoes produce no second event/callback.
- Every remote write costs each subscribed client one extra GET round trip (the no-cid re-pair path).
- Never-loaded slots log the `carried no cid on a never-loaded slot` warning on every frame.

The client degrades *correctly* (no data corruption), but the PR's headline feature needs a companion server change to actually work. **Quoting pitfall for whoever writes that change:** the client records `setResult.etag` verbatim from the `x-cid` header, which the server emits *quoted* — `'x-cid': "\"${newCID}\""` (`routes.ts:1016`). The broadcast `cid` must be that exact quoted string. Broadcasting the bare CID would keep suppression permanently broken, and worse: `_handleRemote` stores the frame's `cid` as the mirror etag (`kv.ts:2293`), so an unquoted cid would make the next local write send an unquoted `if-match` that the server always rejects (`routes.ts:973`) — a guaranteed 412/extra round trip on every write that follows a remote frame.

### Solution

Fix.

corrideat:
Closes #151 

AI disclosure: cid logic produced with Claude Fable 5.

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -77262,7 +77262,7 @@ function registerRoutes(app) {
       }
       const type = cidLookupTable[parsed.code] || "application/octet-stream";
       return c.body(blobOrString, 200, {
-        "ETag": `"${hash3}"`,
+        "ETag": quoteEtag(hash3),
         "Cache-Control": "public,max-age=31536000,immutable",
         // CSP to disable everything -- this only affects direct navigation to the
         // `/file` URL.
--- a/build/main.js
+++ b/build/main.js
@@ -76834,6 +76834,9 @@ function getClientIP(c) {
 function notFoundNoCache(c) {
   return c.body(null, 404, { "Cache-Control": "no-store" });
 }
+function quoteEtag(cid) {
+  return `"${cid}"`;
+}
 function safePathWithin(base2, subpath) {
   const normalizedBase = path5.resolve(base2);
   const resolved = path5.resolve(normalizedBase, subpath);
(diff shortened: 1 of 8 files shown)


corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/30168813312)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/30181164600)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/30182245055)._

corrideat:
Approved

taoeffect (approved):
LGTM!

PR #161, opened 2026-08-17
Support for differently named contracts.

### LINKED ISSUE #160: Support for differently-named identity contracts
### Problem

This is a known but undocumented issue surfaced by <https://github.com/akhileshthite/todomvc/pull/1>:

> The contract must be named `gi.contracts/identity`. POST `/event` rejects an
unattributed first message for any other name, and registers a username only
for that type.

The relevant logic:

```ts
          // Only allow identity contracts to be created without attribution
          if (!credentials?.billableContractID && deserializedHEAD.isFirstMessage) {
            const manifest = await sbp('chelonia.db/get', deserializedHEAD.head.manifest)
            let name: string
            try {
              if (!manifest) throw new Error('empty manifest')
              const parsedManifest = JSON.parse(manifest)
              ;({ name } = JSON.parse(parsedManifest.body))
            } catch (e) {
              if (e instanceof HTTPException) throw e
              throw new HTTPException(422, { message: 'Invalid manifest' })
            }
            if (name !== 'gi.contracts/identity') {
              throw new HTTPException(401, { message: 'This contract type requires ownership information' })
            }
            // ...
          }
```

The rationale for this is / was that identity contracts are special from an accounting perspective, since they aren't owned by any other contract, and they're also 'free' to create. The check is there as sanity check to prevent creating other types of contracts for 'free'.

### Solution

At a very minimum, the `'gi.contracts/identity'` value needs to be configurable, so that different kinds of identity contracts can be used.

An even better solution is to find a way to remove the check entirely while still honouring the intent behind it (free identity contract registration, forced ownership information for other contracts).

corrideat:
Closes #160

AI disclosure: Used GLM-5.3 for planning and code generation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,7 +418,46 @@ email = "admin@example.com"
 
 [database]
 backend = "sqlite"
+
+# Signup size caps and the billing free tier; see
+# "Signup, registration, and billing" below.
+# [server.signup]
+# maxFirstMessageBytes = 5120
+# maxContractSizeBytes = 512000
+#
+# [server.billing]
+# freeAllowanceBytes = 10485760
 ```
+
+#### Signup, registration, and billing
+
+The first message of a contract posted to `POST /event` without shelter
+credentials (an *ownerless* first message) creates a new **billable entity** —
+an identity-like root contract. This is name-agnostic: any contract manifest is
+accepted, not just `gi.contracts/identity`, and username registration via the
+`shelter-namespace-registration` header likewise works for any such contract.
+Contract naming therefore no longer affects registration. Ownerless first
+messages are still subject to:
+
+- `server.signup.disabled` (default `false`) — reject new registrations with
+  `403` when `true`.
+- `server.signup.limit.{minute,hour,day}` (defaults `2`/`10`/`50` per IP) —
+  reject with `429` when an IP exceeds a limit. Disable all rate limiting with
+  `server.signup.limit.disabled = true`.
+- `server.signup.maxFirstMessageBytes` (default `5120`) — reject with `413`
+  when the serialized first message exceeds this size.
+- `server.signup.maxContractSizeBytes` (default `512000`) — reject with `413`
+  when the combined size of the contract sources referenced by the manifest
+  (including the slim variant, if any) exceeds this size.
+
+Each billable entity is billed for its total owned storage: the root contract
+plus everything it owns (e.g. direct-message contracts, file attachments, and
+KV data). Storage up to `server.billing.freeAllowanceBytes` (default
+`10485760`, i.e. 10 MiB, shared across all of the entity's owned resources) is
+free; only the excess is charged. Set it to `0` to disable the free tier and
+charge from the first byte. Because the billing worker runs in a separate
+process without access to `chel.toml`, the server persists the allowance to the
+database at startup, so changes take effect on the next server start.
 
 ### `chelonia.json` — App Properties
 
--- a/src/serve/routes.ts
+++ b/src/serve/routes.ts
@@ -367,20 +367,33 @@ export function registerRoutes (app: Hono): void {
             throw new HTTPException(422, { message: 'Invalid manifest' })
           }
           const credentials = c.get('credentials') as AuthCredentials | undefined
-          // Only allow identity contracts to be created without attribution
+          // Unattributed (ownerless) first messages, i.e. identity contract
+          // registration, are only allowed within configured size limits
           if (!credentials?.billableContractID && deserializedHEAD.isFirstMessage) {
+            if (Buffer.byteLength(payload) > nconf.get('server:signup:maxFirstMessageBytes')) {
+              throw new HTTPException(413, { message: 'First message exceeds size limit' })
+            }
             const manifest = await sbp('chelonia.db/get', deserializedHEAD.head.manifest)
-            let name: string
+            let contractSourceHashes: string[]
             try {
               if (!manifest) throw new Error('empty manifest')
               const parsedManifest = JSON.parse(manifest)
-              ;({ name } = JSON.parse(parsedManifest.body))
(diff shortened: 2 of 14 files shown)


corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/32290274941)._

src/serve/routes.ts
```diff
@@ -367,20 +367,33 @@ export function registerRoutes (app: Hono): void {
...
-            if (name !== 'gi.contracts/identity') {
-              throw new HTTPException(401, { message: 'This contract type requires ownership information' })
+            let contractSizeBytes = 0
+            for (const hash of contractSourceHashes) {
+              const source = await sbp('chelonia.db/get', hash)
+              if (typeof source !== 'string') throw new HTTPException(422, { message: 'Missing contract source' })
+              contractSizeBytes += Buffer.byteLength(source)
+            }
+            if (contractSizeBytes > nconf.get('server:signup:maxContractSizeBytes')) {
+              throw new HTTPException(413, { message: 'Contract source exceeds size limit' })
```
corrideat: These added as per @taoeffect's request. It doesn't seem useful to me to check for `maxContractSizeBytes`, as the contract source size is not a concern here (contract sources are uploaded by a different mechanism).

src/serve/routes.ts
```diff
@@ -367,20 +367,33 @@ export function registerRoutes (app: Hono): void {
...
               throw new HTTPException(422, { message: 'Invalid manifest' })
             }
-            if (name !== 'gi.contracts/identity') {
-              throw new HTTPException(401, { message: 'This contract type requires ownership information' })
+            let contractSizeBytes = 0
+            for (const hash of contractSourceHashes) {
+              const source = await sbp('chelonia.db/get', hash)
+              if (typeof source !== 'string') throw new HTTPException(422, { message: 'Missing contract source' })
+              contractSizeBytes += Buffer.byteLength(source)
+            }
```
akhileshthite: Ordering question. The manifest read and these source reads now happen before
the `signup.disabled` check (line 398) and before the rate limiter below it.

So on a server with `signup.disabled = true`, every request still does one DB
read for the manifest plus up to `maxContractSizeBytes` of source reads, 500 KB
by default, before it returns 403. Same for an IP that is already over the rate
limit, it pays that cost on every attempt.

Before this PR the only work ahead of those two checks was reading the manifest
and pulling the name out of it? Now it is bounded by config, but the bound is
500 KB per request.

Can `signup.disabled` and the rate limiter move above the manifest read? The
`maxFirstMessageBytes` check at line 373 can stay where it is, the payload is
already in memory so it costs nothing.

akhileshthite (changes requested):

src/serve/routes.ts
corrideat: Good point. I think we could move those pre-existing checks to happen before these (like in my other comment, I also think we don't need the contract source checks)

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/33007788392)._

corrideat:
LGTM. README.md issues can be addressed in #164

akhileshthite (approved):
Looks ready, thanks!

src/serve/routes.ts
```diff
@@ -367,37 +299,29 @@ export function registerRoutes (app: Hono): void {
...
-                console.warn('rate limit hit for IP:', ip)
-                throw new HTTPException(429, { message: 'Rate limit exceeded' })
-              }
+            if (!SIGNUP_LIMIT_DISABLED && !await consumeSignupToken(limiters, ip)) {
+              console.warn('rate limit hit for IP:', ip)
+              throw new HTTPException(429, { message: 'Rate limit exceeded' })
             }
+            const manifest = await sbp('chelonia.db/get', deserializedHEAD.head.manifest)
+            const contractSourceHashes = parseContractSourceHashes(manifest)
+            await assertContractSourcesWithinCap(contractSourceHashes, SIGNUP_MAX_CONTRACT_SIZE_BYTES)
```
corrideat: @taoeffect Before you merge this, please look at this.

taoeffect: Looks good to me AFAICT!

README.md
```diff
@@ -418,7 +418,102 @@ email = "admin@example.com"
...
+  1 MiB `POST /event` body limit) nor rate-limited.
+- File uploads are capped per request by `server.fileUploadMaxBytes`, but not in
+  aggregate.
+- Storage beyond the free allowance is charged for and the resulting balance is
+  recorded, but nothing refuses service on account of it.
+
+So a registration that gets through can keep writing. `server.signup.disabled`
+is the only setting that stops new entities from being created at all; bounding
+what existing ones write is, until charge enforcement exists, a job for whatever
+fronts the server (e.g. a reverse proxy).
```
taoeffect: Besides destroying our README.md even further, this junk fills up every single agent context session and therefore:

- Makes agents dumber
- Costs extra $$
- Slows down agents

Please delete all of it or transfer it to an appropriate `docs/` location 🙏

taoeffect (changes requested):
Great work @corrideat! Need a docs update to merge!

src/serve/routes.ts
taoeffect: Could you please move these lines to the `/file` endpoint and comment them out and add a comment to the effort of:

```js
// Uncomment this when we support uploading arbitrary contracts, and only run the check when a contract is being uploaded
```

akhileshthite (approved):
Tested this against our TodoMVC. Renamed our contract from
`gi.contracts/identity` to `todomvc/identity`, dropped the workaround, and ran
the suite against a local build of this branch: all 35 e2e tests pass,
including signup, username registration and login on a browser that has never
seen the account.

The same rename against the published 3.4.0 fails every signup with a 401, so
the tests really are exercising this change and not just passing anyway.

PR #162, opened 2026-08-19
Use `better-sqlite3` instead of `@db/sqlite`

### LINKED ISSUE #150: Investigate better-sqlite3 backend
### Problem

We used to use `better-sqlite3`, but switched to `jsr:@db/sqlite` after we found it wasn't compatible with Deno.

Now apparently [`better-sqlite3` has been rewritten to be compatible with Deno](https://github.com/WiseLibs/better-sqlite3/issues/1205#event-28253493093).

@corrideat says that he's "not thrilled" with some aspects of `jsr:@db/sqlite`:

> With that said, I don't like the install part of what we use in Deno. At runtime, it'll try to fetch and download the sqlite dynamic library. Ideally, this wouldn't happen.

> Other than remote code fetching and that the library fetched may not work on all systems (had a problem with the glibc version it used), I haven't seen any issues with what we have. (It is possible to override the library being used)

So it's not really a problem, just something to research.

### Solution

Investigate if it's worth switching back to `better-sqlite3` now that it works with Deno.

If it doesn't download sqlite from the internet and uses the one bundled with the OS that would be preferable.

Explain your reasoning in a comment below.

corrideat:
Closes #150

Before merging this PR, we need to ensure that tests pass on GI (using the SQLite backend)

<del>A known issue with this PR as it stands right now is that the bundle will contain all prebuilt SQLite binaries (e.g., the Linux bundle will include Windows code and vice versa). This is not directly harmful but makes the bundle larger than it needs to be.</del>

AI disclosure: Used Opus 5 for code generation and planning


corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/32285826951)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/32433011793)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/32438618645)._

src/serve/database-sqlite.ts
```diff
@@ -1,20 +1,76 @@
...
+// 'Cannot find module ...better_sqlite3.node', or a prebuilds/ path reported
+// without a standard code) does name it.
+export function rewordedAddonLoadError (cause: unknown): Error | null {
+  const message = cause instanceof Error ? cause.message : ''
+  if (!/better_sqlite3\.node|prebuilds[\\/]/.test(message)) return null
+  return new Error(
+    'Could not load the SQLite native addon. The pre-built binaries chel ships ' +
+    'are linked against glibc, so musl-based distributions (Alpine, for ' +
+    'example) are not covered; run chel from source with Deno there, or pick a ' +
+    'different database backend.',
```
akhileshthite: Small doubt about the advice in this message, not the code.

It tells musl users to run chel from source with Deno instead. But on musl
there is no matching prebuild for better-sqlite3 either, so Deno would have to
build the addon with node-gyp, and Deno does not run npm lifecycle scripts
unless you pass `--allow-scripts`, plus the machine needs python and a
compiler.

So does that path actually work on Alpine as written? If it needs
`--allow-scripts` and build tools, maybe say so here and in the README (line
490 makes the same claim), otherwise we send people to a second thing that also
fails. If it does work out of the box then ignore me, I have not tried it on
Alpine.

corrideat: Needing build tools is something that _could_ be mentioned, although it's IMHO implied if we're directing users to building from source.

The larger question here is that, currently, it doesn't easily work on Alpine without some patching due to https://github.com/denoland/deno/issues/33948.

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/chel/actions/runs/32998031834)._

akhileshthite (dismissed):

corrideat:
LGTM

taoeffect:
Let me know @corrideat when you've tested this with Group Income and I can merge (if all looks good with the PR)

.github/workflows/ci.yml
```diff
@@ -17,6 +17,35 @@ jobs:
...
+    - uses: denoland/setup-deno@e95548e56dfa95d4e1a28d6f422fafe75c4c26fb
+      with:
+        deno-version: v2.9.5
+    # The compile embeds files straight out of node_modules, so the addon has
+    # to be materialised first; the committed bundle in build/ is what gets
+    # compiled, so nothing else needs building.
+    - name: Install dependencies
+      run: deno install
+    - name: Compile the host binary and load its SQLite addon
+      run: CHEL_SMOKE_COMPILE=1 deno task smoke
```
taoeffect, line 45: What is this smoke thing? Is it needed?

Is it an improvement to the test coverage (i.e. it's useful regardless of whether we use `better-sqlite3` instead of `@db/sqlite`?

EDIT: based on a discussion on a video call it sounded like this is an improvement to verify that the `chel` binary successfully builds and runs, but the comment doesn't say that. Instead it's a bunch of Claudish gibberish. If I understand correctly, please manually edit the comment to be clear.

We don't have it yet, but consider it a non-spoken rule that Claude is not allowed to write any documentation for us because it cannot speak English properly. You can, if you want, use some other model that knows how to speak English.

test/compile-smoke.test.ts
```diff
@@ -0,0 +1,58 @@
+// Opt-in end-to-end check that a compiled binary can actually load the single
+// SQLite addon it embeds.
+//
+// This is the only test that exercises the prebuild mapping in
+// scripts/paths.ts for real: everything else compares strings. A wrong mapping
+// produces a binary that compiles and starts fine and only fails the moment
+// the sqlite backend is used, so it is worth having, but a compile costs
+// minutes and ~100 MB of disk. It therefore stays out of `deno task test` and
+// runs on request:
+//
```
corrideat: Rewrite prose

README.md
```diff
@@ -590,13 +611,25 @@ gitignored `dist/` directory. `deno task dist` remains usable on its own, at
...
+because the binaries no longer share their inputs: each embeds only its own
+platform's SQLite addon instead of all eight, which is worth 14.8 MB per
+binary (~7.5 MB per compressed tarball). A practical side effect is that
+reinstalling the SQLite package only invalidates the targets whose own addon
+actually changed.
+
+To verify that a freshly compiled binary really can load the addon it embeds,
+run `CHEL_SMOKE_COMPILE=1 deno task smoke`. It compiles the host platform's
+binary and runs a real SQLite migration through it. It is excluded from
+`deno task test` because it costs minutes and ~100 MB.
```
taoeffect: None of this belongs in the README.

1. It is poorly written (because Claude wrote it)
2. It's details that don't belong in the README

AGENTS.md
```diff
@@ -88,14 +96,42 @@ scripts/
...
+  `COMPILE_FLAGS` / `NATIVE_ADDON_PATHS` constants, and each binary embeds only
+  its own platform's prebuilt SQLite addon.
+- Cache keys follow suit: `targetFingerprint(target)` in `scripts/binaries.ts`
+  replaces `bundleFingerprint()`, and `tarFingerprint()` in `scripts/compile.ts`
+  now takes the target too.
+- `ALL_NATIVE_ADDON_PATHS` is the union over all targets; it is what the mtime
+  normalisation (needed for reproducible binaries) has to walk.
+- `CHEL_SMOKE_COMPILE=1 deno task smoke` is the only test that actually
+  compiles a binary and executes it; it is skipped by `deno task test`.
+
```
taoeffect: > [!WARNING]
> **Written by Claude — an AI that doesn't know English**

Please manually review this and either remove or shorten it to just the relevant highlights.

corrideat: Simplified this. Is it better?

README.md
corrideat: Simplified.

deno.json
```diff
@@ -27,7 +28,12 @@
     "jsr:@std/path/": "jsr:@std/path@1.1.1/",
     "jsr:@std/streams/": "jsr:@std/streams@1.0.10/",
     "jsr:@std/io/": "jsr:@std/io@0.225.2/",
-    "jsr:@db/sqlite": "jsr:@db/sqlite@0.13.0",
+    "npm:better-sqlite3": "npm:better-sqlite3@13.0.3",
+    // TODO: re-pin once DefinitelyTyped publishes typings for better-sqlite3
+    // v10+. 9.6.0 is currently the newest one there is, so `deno check` only
+    // knows the v9-era API of the v13 driver above; see the header comment in
+    // src/serve/database-sqlite.ts.
+    "npm:@types/better-sqlite3": "npm:@types/better-sqlite3@9.6.0",
```
taoeffect, line 36: - What is this comment? It says "re-pin" — what does that mean?
- Does deno.json support comments?
- If we're using v13 of better-sqlite3 why is it using types for better-sqlite3 v9? Wouldn't they be radically different?

deno.json
```diff
@@ -43,6 +49,10 @@
     "npm:three": "npm:three@0.151.3",
     "npm:turtledash": "npm:turtledash@1.0.3",
     "npm:vue": "npm:vue@2.7.16",
+    // Bare specifier for src/types/vue-shims.d.ts, whose `declare module`
+    // bodies import 'vue'; resolved from node_modules by editors using
+    // tsconfig.json and by `deno check` via this mapping.
+    "vue": "npm:vue@2.7.16",
```
taoeffect: I don't understand what this comment is saying. It sounds like Claude wrote it. Could you please nuke it from outerspace?

README.md
```diff
@@ -589,14 +610,14 @@ the tarballs and the npm sub-packages share one set of them, kept under the
...
-To rebuild everything from scratch anyway, set `CHEL_FORCE_COMPILE=1` or delete
-`dist/`.
+The cache operates on content, not on timestamps. When the bundle, the Deno
+version, or a compile flag changes, `deno task dist` compiles the binaries
+again. When nothing changes, it does no compile work. To compile all binaries
+again, set `CHEL_FORCE_COMPILE=1` or delete `dist/`.
+
+To check that a compiled binary loads its SQLite addon, run
+`CHEL_SMOKE_COMPILE=1 deno task smoke`. This test compiles a binary and is
+therefore not part of `deno task test`.
```
taoeffect, line 620: None of this, and most of the README, belongs in the README. I opened #164 to address this. Do you want to address it now or include it and deal with it in a followup?

AGENTS.md
taoeffect: It seems like too many details to be in AGENTS.md tbh. Why is it mentioning specific function names? These details belong in comments in the relevant files. AGENTS.md is supposed to be:

- A TOC for the project
- Guidelines for how to work with the project

And that's it.

deno.json
corrideat: This might be an example of what you refer to as Claudish, yet it's perfectly readable (and clear) to me.

1. Since we're pinning a version (`9.6.0` in this case), it'll need to be pinned again (aka, 're-pin'-ed) "once DefinitelyTyped publishes [updated] typings for better-sqlite3"
2. Apparently so. The official docs quote it supports `json` and `jsonc`. So far, this hasn't broken anything.
3. They might radically different types, but as the comment explains ('9.6.0 is currently the newest one there is'), it's the best we have. See <https://www.npmjs.com/package/@types/better-sqlite3?activeTab=versions>

taoeffect: I see. That could be restated as, "Update the pinned version here to v10 once ...." — that would be much clearer. 🙏 

> They might radically different types, but as the comment explains ('9.6.0 is currently the newest one there is'), it's the best we have

I'm not a TypeScript expert, so this is news to me that you can use wrong/bad types and things be OK. Wouldn't using the wrong types be *worse* than using no types at all?

README.md
corrideat: I think we can deal with this in #164, so that we do it all at once.

deno.json
corrideat: It's saying `src/types/vue-shims.d.ts` uses `'vue'` instead of `'npm:vue'`, and therefore we need to pin `vue` like we do `npm:vue`.

taoeffect: Could you please update the comment then to say that?

AGENTS.md
corrideat: Simplified `AGENTS.md` some.

taoeffect (approved):

